PR #3, opened 2020-10-09
FTUP feature implementation

sergeymatov:
This is the implementation of F-TEID allocation in the UP function feature. Based on https://itectec.com/spec/5-5-f-teid-allocation-and-release/

### DIFF
--- a/upf/pfcp.c
+++ b/upf/pfcp.c
@@ -794,60 +794,62 @@ decode_f_teid (u8 * data, u16 length, void *p)
 {
   pfcp_f_teid_t *v = p;
 
-  if (length < 5)
-    return PFCP_CAUSE_INVALID_LENGTH;
+  if (length < 1)
+   return PFCP_CAUSE_INVALID_LENGTH;
 
   v->flags = get_u8 (data) & 0x0f;
-  v->teid = get_u32 (data);
-  length -= 5;
 
-  if (v->flags & F_TEID_CH)
-    {
-      if (v->flags & (F_TEID_V4 | F_TEID_V6))
-	{
-	  pfcp_debug ("PFCP: F-TEID with invalid flags (CH and v4/v6): %02x.",
-		      v->flags);
-	  return -1;
-	}
-    }
-  else
-    {
-      if (v->flags & F_TEID_CHID)
-	{
-	  pfcp_debug
-	    ("PFCP: F-TEID with invalid flags (CHID without CH): %02x.",
-	     v->flags);
-	  return -1;
-	}
-      if (!(v->flags & (F_TEID_V4 | F_TEID_V6)))
-	{
-	  pfcp_debug ("PFCP: F-TEID without v4/v6 address: %02x.", v->flags);
-	  return -1;
-	}
-    }
-
-  if (v->flags & F_TEID_V4)
+  if (!(v->flags & F_TEID_CH))
     {
       if (length < 4)
-	return PFCP_CAUSE_INVALID_LENGTH;
+        return PFCP_CAUSE_INVALID_LENGTH;
 
-      get_ip4 (v->ip4, data);
+      v->teid = get_u32 (data);
       length -= 4;
+
+      if (v->flags & F_TEID_CHID)
+        {
+          pfcp_debug
+            ("PFCP: F-TEID with invalid flags (CHID without CH): %02x.",
+             v->flags);
+          return -1;
+        }
+
+      if (!(v->flags & (F_TEID_V4 | F_TEID_V6)))
+        {
+          pfcp_debug ("PFCP: F-TEID without v4/v6 address: %02x.", v->flags);
+          return -1;
+        }
+
+        if (v->flags & F_TEID_V4)
+          {
+            if (length < 4)
+              return PFCP_CAUSE_INVALID_LENGTH;
+
+            get_ip4 (v->ip4, data);
+            length -= 4;
+          }
+
+        if (v->flags & F_TEID_V6)
+          {
+            if (length < 16)
+              return PFCP_CAUSE_INVALID_LENGTH;
+
+            get_ip6 (v->ip6, data);
+            length -= 16;
+          }
+
     }
-
-  if (v->flags & F_TEID_V6)
-    {
-      if (length < 16)
-	return PFCP_CAUSE_INVALID_LENGTH;
-
-      get_ip6 (v->ip6, data);
-      length -= 16;
-    }
-
-  if (v->flags & F_TEID_CHID)
+  else // CH flag found
     {
       if (length < 1)
 	return PFCP_CAUSE_INVALID_LENGTH;
+
+      if (!(v->flags & F_TEID_V6) && !(v->flags & F_TEID_V4))
+        {
+          pfcp_debug ("PFCP: F-TEID with CH flag should have at least v4/v6 flag:%02x", v->flags);
+          return -1;
+        }
 
       v->choose_id = get_u8 (data);
     }
--- a/upf/upf.h
+++ b/upf/upf.h
@@ -626,6 +626,9 @@ typedef struct
   /* DPO locks */
   u32 dpo_locks;
 
+  /* TEID table by choose_id */
+  u32 *teid_by_chid;
+
   f64 unix_time_start;
 } upf_session_t;
 
@@ -819,6 +822,10 @@ typedef struct
   uword *upf_app_by_name;
   /* adf apps vector */
   upf_adf_app_t *upf_apps;
+
+  //TODO: Change to UPF flags?
+  u32 pfcp_spec_version;
+  u32 rand_base;
 } upf_main_t;
 
 extern const fib_node_vft_t upf_vft;
--- a/upf/upf_pfcp_api.c
+++ b/upf/upf_pfcp_api.c
@@ -32,6 +32,8 @@
 #include <vppinfra/types.h>
 #include <vppinfra/vec.h>
 #include <vppinfra/format.h>
+#include <vppinfra/random.h>
+#include <vppinfra/sparse_vec.h>
 #include <vnet/fib/ip4_fib.h>
 #include <vnet/fib/ip6_fib.h>
 #include <vnet/ip/ip6_hop_by_hop.h>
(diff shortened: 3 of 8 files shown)


upf/upf.h
```diff
@@ -819,6 +826,11 @@ typedef struct
   uword *upf_app_by_name;
   /* adf apps vector */
   upf_adf_app_t *upf_apps;
+
+  //TODO: Change to UPF flags?
+  u32 pfcp_spec_version;
+  u32 rand_base;
+  u32 choosed_teid[UPF_CHOOSEID_MAX];
```
RoadRunnr, line 833: this is not needed and not used

upf/upf.h
```diff
@@ -626,6 +630,9 @@ typedef struct
   /* DPO locks */
   u32 dpo_locks;
 
+  /* TEID table by choose_id */
+  u32 *teid_by_choose_id_table;
```
RoadRunnr, line 634: what do think about renaming this to `teid_by_chid` ??

* `chid` is IMHO clear enough
* a `_table` suffix is not used in VPP or UPG for any of the lookup tables, the usual naming is just `x_by_y`

Also, what about changing all other instance of `choose_id` to `chid` ??

upf/upf_pfcp_api.c
```diff
@@ -803,16 +808,140 @@ lookup_nwi (u8 * name)
   return pool_elt_at_index (gtm->nwis, p[0]);
 }
 
+static u32
+generate_teid (u32 start, u32 mask)
+{
+  return random_u32(&start) & (UINT32_MAX ^ mask);
```
RoadRunnr, line 814: 0 and ~0 (0xffffffff) are reserved TEIDs, we need to make sure we never generate those ids.

upf/upf_pfcp_api.c
```diff
@@ -803,16 +808,140 @@ lookup_nwi (u8 * name)
...
+    }
+/*  if (add)
+  {
+    hash_set (sx->teid_by_choose_id_table, choose_id, teid);
+  } else
+  {
+    hash_unset(sx->teid_by_choose_id_table, teid);
+  }
+*/
+}
```
RoadRunnr, line 872: this is a very non VPP way to use this
```suggestion
{
  u32 *n;

  if (add)
    {
      n = sparse_vec_validate (sx->teid_by_choose_id_table, choose_id);
      n[0] = teid;
    }
  else
    {
      n = sparse_vec_validate (sx->teid_by_choose_id_table, choose_id);
      n[0] = 0;
    }
}
```

upf/upf_pfcp_api.c
```diff
@@ -1041,27 +1246,89 @@ handle_update_pdr (upf_session_t * sx, pfcp_update_pdr_t * update_pdr,
...
+
+        if (!ISSET_BIT (pdr->pdi.grp.fields, PDI_SDF_FILTER) &&
+            !ISSET_BIT (pdr->pdi.grp.fields, PDI_APPLICATION_ID))
+          {
+            /* neither SDF, nor Application Id, generate a wildcard
+               ACL to make ACL scanning simpler */
+            update->pdi.fields |= F_PDI_SDF_FILTER;
+            vec_reset_length (update->pdi.acl);
+            vec_add1 (update->pdi.acl, wildcard_acl);
+          }
```
RoadRunnr, line 1262: moving this code block around makes the change harder to read. It could be just a diff fragment, but could you try to keep the block in place?

RoadRunnr (changes requested):

mgumz:
@sergeymatov what is "FTUP" about? this PR deserves a bit more context

sergeymatov:
@mgumz FTUP is for F_TEID allocation in UP function, I will rename the PR to be more clear

upf/upf_pfcp_api.c
sergeymatov: I've lifted this code block up to have UE IP information already parsed so I could then fill created_pdrs response with both fteid and UE IP.

upf/upf_pfcp_api.c
```diff
@@ -1041,27 +1246,89 @@ handle_update_pdr (upf_session_t * sx, pfcp_update_pdr_t * update_pdr,
...
+              {
+                update->pdi.teid.ip4 = res->ip4;
+                update->pdi.teid.flags |= F_TEID_V4;
+              }
+            if ((!is_zero_ip6_address (&res->ip6)) && (pdr->pdi.f_teid.flags && F_TEID_V4))
+              {
+                update->pdi.teid.flags |= F_TEID_V6;
+                update->pdi.teid.ip6 = res->ip6;
+              }
+            pfcp_created_pdr_t *created_pdr;
```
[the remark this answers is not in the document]
sergeymatov: I suppose it's better to move this on top of the code block with F_TEID field checks. This is only needed if we are filling up response with created_pdrs if we done teid generation on UP side. This will be unused in all other situations

upf/upf_pfcp_api.c
```diff
@@ -803,16 +808,140 @@ lookup_nwi (u8 * name)
...
+        }
+      gtm->rand_base = teid;
+      if ((flags & F_TEID_CHID) && choose_id)
+        {
+          teid_add_del_by_choose_id(sx, choose_id, teid, 1 /*add*/);
+        }
+   }
+
+  return teid;
+}
```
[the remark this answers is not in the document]
sergeymatov: agree. Just adding (!added) to condition for ipv6 should be fine

upf/upf_pfcp_api.c
```diff
@@ -857,11 +1075,41 @@ handle_create_pdr (upf_session_t * sx, pfcp_create_pdr_t * create_pdr,
 	create->pdi.nwi_index = nwi - gtm->nwis;
       }
 
+    p = hash_get_mem (gtm->nwi_index_by_name, nwi->name);
```
RoadRunnr, line 1078: nwi could be `NULL`

upf/upf_pfcp_api.c
```diff
@@ -857,11 +1075,41 @@ handle_create_pdr (upf_session_t * sx, pfcp_create_pdr_t * create_pdr,
...
+    if (!p)
+      {
+	upf_debug("Can't find nwi idx by name\n");
+        r = -1;
+	break;
+      }
+
+    idx = p[0];
+    res = &gtm->upip_res[idx];
```
RoadRunnr, line 1087: wrong, you need to lookup `res` on `upip_res_index` using the nwi index (which be calculates as `nwi - gtp-nwis`).

Although, I have to admit that storing the UPIP information in a pool is a bit over complicating things. A simple vector that uses the same index as the NWI would be much simpler. Alternatively, storing the UPIP information directly in the NWI struct would also work nicely.
Anyhow, I you want to change it, do it in a separate commit before the FTUP changes.

upf/upf_pfcp_api.c
RoadRunnr: UE IP is only part of `Created PDR` when UP IP assignment by UP function was requested. Since we don't support that feature (yet), including the UE IP is not needed.

upf/upf_pfcp_api.c
```diff
@@ -803,9 +808,217 @@ lookup_nwi (u8 * name)
...
+        }
+
+      vec_add2 (*created_pdr_vec, created_pdr, 1);
+      memset (created_pdr, 0, sizeof (*created_pdr));
+      created_pdr->pdr_id = process_pdr->id;
+      created_pdr->f_teid = process_pdr->pdi.teid;
+      created_pdr->ue_ip_address = process_pdr->pdi.ue_addr;
+      SET_BIT(created_pdr->grp.fields, CREATED_PDR_PDR_ID);
+      SET_BIT(created_pdr->grp.fields, CREATED_PDR_F_TEID);
+      SET_BIT(created_pdr->grp.fields, CREATED_PDR_UE_IP_ADDRESS);
```
RoadRunnr, line 1004: not needed, see later comment and UP function based IP assignment

RoadRunnr (changes requested):
please review the commit yourself for white space and formatting

* no double newlines
* should (mostly) pass VPP checkstyle.sh script

upf/pfcp.c
```diff
@@ -794,54 +794,50 @@ decode_f_teid (u8 * data, u16 length, void *p)
...
+
+            get_ip6 (v->ip6, data);
+            length -= 16;
+          }
 
-      get_ip6 (v->ip6, data);
-      length -= 16;
     }
 
   if (v->flags & F_TEID_CHID)
```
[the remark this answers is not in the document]
sergeymatov: hmm, am I getting this wrong?

– Bit 1 – V4: If this bit is set to "1" and the CH bit is not set, then the IPv4 address field shall be present, otherwise the IPv4 address field shall not be present.

– Bit 2 – V6: If this bit is set to "1" and the CH bit is not set, then the IPv6 address field shall be present, otherwise the IPv6 address field shall not be present.

In old version, if CH is set flags v4/v6 can't be present in IE_FTEID. However, in this case if CH is set one/both of version flags can be present. From my point of view the only check we can plant here looks like this: if CH is set and one/both version flags are set, there should be no IP(4/6) field present in a IE_FTEID. I will put this check.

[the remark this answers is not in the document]
RoadRunnr: If CH flag is set, the v4/v6 field can not be present, however the flags can be set. In fact, independent of who CH ist set, at least on the v4/v6 flags need to be set.

TS 29.244, V16.4.0 has this:

> At least one of the V4 and V6 flags shall be set to "1", and both may be set to "1" for both scenarios:
> -	when the CP function is providing F-TEID, i.e. both IPv4 address field and IPv6 address field may be present; or
> -	when the UP function is requested to allocate the F-TEID, i.e. when CHOOSE bit is set to "1", and the IPv4 address and IPv6 address fields are not present.

upf/upf_pfcp_api.c
```diff
@@ -803,9 +809,210 @@ lookup_nwi (u8 * name)
...
+  gtm->rand_base = teid;
+  if ((flags & F_TEID_CHID))
+    {
+      u32 *n;
+      n = sparse_vec_validate (sx->teid_by_chid, chid);
+      n[0] = teid;
+    }
+
+  return teid;
+}
```
RoadRunnr, line 937: too complicated, untested

```suggestion
static u32
process_teid_generation (upf_main_t *gtm, u8 chid, u32 flags,
		  upf_upip_res_t *res, upf_session_t *sx)
{
  u8 retry_cnt = 10;
  u32 teid = 0;
  bool ok;


  do
    {
      teid = generate_teid(gtm->rand_base, res->mask);
      gtm->rand_base = teid + retry_cnt;

      if (!validate_teid(teid))
	{
	  retry_cnt--;
	  continue;
	}

      ok =
	(!(flags & F_TEID_V4) || teid_v4_lookup(gtm, teid, res) == 0) &&
	(!(flags & F_TEID_V6) || teid_v6_lookup(gtm, teid, res) == 0);

      if (ok)
	break;

    }
  while (retry_cnt > 0);

  if (!ok)
    /* can't generate unique for given UPIP resource */
    return 0;

  if ((flags & F_TEID_CHID))
    {
      u32 *n = sparse_vec_validate (sx->teid_by_chid, chid);
      n[0] = teid;
    }

  return teid;
}
```

upf/upf_pfcp_api.c
```diff
@@ -803,9 +809,210 @@ lookup_nwi (u8 * name)
...
+  if (!validate_teid(c))
+    {
+      //we are doing single try to regenerate, making next
+      //based on generated value
+      ret_cnt += 1;
+      goto retry;
+    }
+
+  return c;
+}
```
RoadRunnr, line 839: this could also be moved into process_teid_generation

sergeymatov: Thanks Andreas. I will apply changes requested right after we will deal with PR#5 :)

RoadRunnr:
@sergeymatov this PR is superseded by #17, right? If so, please close this one.

sergeymatov:
duplicate of #17

PR #7, opened 2020-10-15
Fix stopping PFCP timers

ivan4th:
There was a corner case of timers "firing" after being stopped because
their "user handles" being already collected into the 'expires' vector
in pfcp_process(). This caused msg_pool corruption upon a node
association being released after a heartbeat timeout, for example.

### DIFF
--- a/upf/upf_pfcp_server.c
+++ b/upf/upf_pfcp_server.c
@@ -290,7 +290,7 @@ upf_pfcp_server_rx_msg (pfcp_msg_t * msg)
 
 	req = pfcp_msg_pool_elt_at_index (psm, p[0]);
 	hash_unset (psm->request_q, msg->seq_no);
-	upf_pfcp_server_stop_timer (req->timer);
+	upf_pfcp_server_stop_msg_timer (req);
 
 	msg->node = req->node;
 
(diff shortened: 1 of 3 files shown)


upf/upf_pfcp_server.c
```diff
@@ -1040,11 +1040,30 @@ static void upf_validate_session_timers ()
...
+void upf_pfcp_server_stop_msg_timer (pfcp_msg_t * msg)
 {
+  if (msg->timer == ~0)
+    return;
+
   pfcp_server_main_t *psm = &pfcp_server_main;
+  u32 id_t1 = ((0x80 | PFCP_SERVER_T1) << 24) | msg->seq_no;
+  u32 id_resp =
+    ((0x80 | PFCP_SERVER_RESPONSE) << 24) | pfcp_msg_get_index (psm, msg);
+  u32 *exp_id;
```
RoadRunnr, line 1052: personal, I don't this style. I would first declare the variable, the check msg->timer and the initialize the variable.

```suggestion
  pfcp_server_main_t *psm = &pfcp_server_main;
  u32 id_resp, *exp_id, id_t1;

    if (msg->timer == ~0)
    return;

  id_t1 = ((0x80 | PFCP_SERVER_T1) << 24) | msg->seq_no;
  id_resp =
    ((0x80 | PFCP_SERVER_RESPONSE) << 24) | pfcp_msg_get_index (psm, msg);
```

ivan4th: Okay, will fix. Did too much C++ and Go in my time :)

ivan4th: Fixed

ivan4th:
Merging, with a followup pending for better approach.

PR #12, opened 2020-10-16
Fix releasing PFCP associations

ivan4th:
This fixes expired timer conflict in a cleaner way than #7, reverting
previous PFCP timer changes.


upf/upf_pfcp_api.c
```diff
@@ -524,7 +524,7 @@ handle_association_setup_request (pfcp_msg_t * req,
        * PFCP Association Setup Response message.
        *
        */
-      pfcp_release_association (n);
+      upf_pfcp_server_deferred_release_association (n - gtm->nodes);
```
RoadRunnr, line 527: Rel. 16 changed the rules on this.

**just a heads up, don't change it now!**

RoadRunnr (approved):
LGTM

upf/upf_pfcp_api.c
ivan4th: ✓
Thanks

PR #17, opened 2020-10-26
FTUP feature implementation

sergeymatov:
This commit enables F_TEID generation in User Plane Function.
Based on 3GPP 29.244 Specification release 16, chapter 5.5

### DIFF
--- a/upf/pfcp.c
+++ b/upf/pfcp.c
@@ -770,21 +770,20 @@ static u8 *
 format_f_teid (u8 * s, va_list * args)
 {
   pfcp_f_teid_t *v = va_arg (*args, pfcp_f_teid_t *);
+  u16 flags = (v->flags & 0xf);
+
+  s = format (s, "CH:%d,CHID:%d,V4:%d,V6:%d,", !!(flags & F_TEID_CH),
+	      !!(flags & F_TEID_CHID), !!(flags & F_TEID_V4),
+	      !!(flags & F_TEID_V6));
 
   if ((v->flags & 0xf) == F_TEID_V4)
-    s = format (s, "%d,IPv4:%U", v->teid, format_ip4_address, &v->ip4);
+    s = format (s, "TEID:%d,IPv4:%U", v->teid, format_ip4_address, &v->ip4);
   else if ((v->flags & 0xf) == F_TEID_V6)
-    s = format (s, "%d,IPv6:%U", v->teid, format_ip6_address, &v->ip6);
+    s = format (s, "TEID:%d,IPv6:%U", v->teid, format_ip6_address, &v->ip6);
   else if ((v->flags & 0xf) == (F_TEID_V4 | F_TEID_V6))
-    s = format (s, "%d,IPv4:%U,IPv6:%U",
+    s = format (s, "TEID:%d,IPv4:%U,IPv6:%U",
 		v->teid, format_ip4_address, &v->ip4,
 		format_ip6_address, &v->ip6);
-  else if ((v->flags & 0xf) == F_TEID_CH)
-    s = format (s, "%d,CH:1", v->teid);
-  else if ((v->flags & 0xf) == (F_TEID_CH | F_TEID_CHID))
-    s = format (s, "%d,CH:1,CHID:%d", v->teid, v->choose_id);
-  else
-    s = format (s, "invalid flags: %02x", v->flags);
 
   return s;
 }
--- a/upf/pfcp.c
+++ b/upf/pfcp.c
@@ -794,24 +794,19 @@ decode_f_teid (u8 * data, u16 length, void *p)
 {
   pfcp_f_teid_t *v = p;
 
-  if (length < 5)
+  if (length < 1)
     return PFCP_CAUSE_INVALID_LENGTH;
 
   v->flags = get_u8 (data) & 0x0f;
-  v->teid = get_u32 (data);
-  length -= 5;
 
-  if (v->flags & F_TEID_CH)
+  if (!(v->flags & F_TEID_CH))
     {
-      if (v->flags & (F_TEID_V4 | F_TEID_V6))
-	{
-	  pfcp_debug ("PFCP: F-TEID with invalid flags (CH and v4/v6): %02x.",
-		      v->flags);
-	  return -1;
-	}
-    }
-  else
-    {
+      if (length < 4)
+	return PFCP_CAUSE_INVALID_LENGTH;
+
+      v->teid = get_u32 (data);
+      length -= 4;
+
       if (v->flags & F_TEID_CHID)
 	{
 	  pfcp_debug
@@ -819,35 +814,44 @@ decode_f_teid (u8 * data, u16 length, void *p)
 	     v->flags);
 	  return -1;
 	}
+
       if (!(v->flags & (F_TEID_V4 | F_TEID_V6)))
 	{
 	  pfcp_debug ("PFCP: F-TEID without v4/v6 address: %02x.", v->flags);
 	  return -1;
 	}
+
+      if (v->flags & F_TEID_V4)
+	{
+	  if (length < 4)
+	    return PFCP_CAUSE_INVALID_LENGTH;
+
+	  get_ip4 (v->ip4, data);
+	  length -= 4;
+	}
+
+      if (v->flags & F_TEID_V6)
+	{
+	  if (length < 16)
+	    return PFCP_CAUSE_INVALID_LENGTH;
+
+	  get_ip6 (v->ip6, data);
+	  length -= 16;
+	}
+
     }
-
-  if (v->flags & F_TEID_V4)
-    {
-      if (length < 4)
-	return PFCP_CAUSE_INVALID_LENGTH;
-
-      get_ip4 (v->ip4, data);
-      length -= 4;
-    }
-
-  if (v->flags & F_TEID_V6)
-    {
-      if (length < 16)
-	return PFCP_CAUSE_INVALID_LENGTH;
-
-      get_ip6 (v->ip6, data);
-      length -= 16;
-    }
-
-  if (v->flags & F_TEID_CHID)
+  else
     {
       if (length < 1)
 	return PFCP_CAUSE_INVALID_LENGTH;
+
+      if (!(v->flags & F_TEID_V6) && !(v->flags & F_TEID_V4))
+	{
+	  pfcp_debug
+	    ("PFCP: F-TEID with CH flag should have at least v4/v6 flag:%02x",
+	     v->flags);
+	  return -1;
+	}
 
       v->choose_id = get_u8 (data);
     }
--- a/upf/test/test_upf.py
+++ b/upf/test/test_upf.py
@@ -102,6 +102,7 @@ class IPv4Mixin(object):
             "upf nwi name cp vrf 0",
             "upf nwi name epc vrf 100",
             "upf nwi name sgi vrf 200",
+            "upf specification release 16",
             "upf pfcp endpoint ip %s vrf 0" % cls.if_cp.local_ip4,
             "ip route add 0.0.0.0/0 table 200 via %s %s" %
             (cls.if_sgi.remote_ip4, cls.if_sgi.name),
@@ -139,6 +140,9 @@ class IPv4Mixin(object):
 
     def ie_fteid(self):
         return IE_FTEID(V4=1, TEID=self.UNODE_TEID, ipv4=self.if_grx.local_ip4)
+
+    def ie_fteid_ch(self):
+        return IE_FTEID(CH=1, CHID=1, choose_id=200, V4=1)
 
     def ie_ue_ip_address(self, SD=0):
         return IE_UE_IP_Address(ipv4=self.ue_ip, V4=1, SD=SD)
@@ -227,6 +231,7 @@ class IPv6Mixin(object):
             "upf nwi name cp vrf 0",
             "upf nwi name epc vrf 100",
             "upf nwi name sgi vrf 200",
+            "upf specification release 16",
             "upf pfcp endpoint ip %s vrf 0" % cls.if_cp.local_ip6,
             "ip route add ::/0 table 200 via %s %s" %
             (cls.if_sgi.remote_ip6, cls.if_sgi.name),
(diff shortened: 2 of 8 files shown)


upf/test/test_upf.py
```diff
@@ -423,6 +431,10 @@ def establish_session(self):
         self.assertEqual(CauseValues[resp[IE_Cause].cause], "Request accepted")
         self.verify_ie_fseid(resp[IE_FSEID])
         self.assertEqual(resp[IE_FSEID].seid, self.cur_seid)
+        if IE_CreatedPDR in resp:
+            self.teid = resp[IE_CreatedPDR][IE_FTEID].TEID
+            self.logger.info("SMATOV: teid ")
```
ivan4th, line 436: Redundant debug print

upf/test/test_upf.py
```diff
@@ -1048,6 +1061,81 @@ def verify_gtp_no_echo(self, to_send):
...
+        self.assertEqual(pkt[l4proto].dport, remote_port)
+        if payload is not None:
+            self.assertEqual(pkt[Raw].load, payload)
+
+    def assert_packet_sent_to_ue(self, payload=None, l4proto=UDP,
+                                  ue_port=12345, remote_port=23456,
+                                  remote_ip=None, expected_seq=42):
+        pkt = self.if_grx.get_capture(1)[0]
+        self.logger.info("SMATOV:")
+        self.logger.info(pkt.show2())
```
ivan4th, line 1109: Redundant debug print

upf/test/test_upf.py
```diff
@@ -1048,6 +1061,81 @@ def verify_gtp_no_echo(self, to_send):
         self.send_from_grx(to_send)
         self.if_grx.assert_nothing_captured()
 
+    def send_from_sgi_to_ue(self, payload=None, l4proto=UDP, ue_port=12345,
```
ivan4th, line 1064: I believe this method only differs `PFCPHelper.send_from_sgi_to_access()` by its name? Then, I guess the right thing to do is to just rename `PFCPHelper.send_from_sgi_to_access()` to `PFCPHelper.send_from_sgi_to_ue()` or `PFCPHelper.send_from_sgi()` as basically they do the same thing

upf/test/test_upf.py
```diff
@@ -1048,6 +1061,81 @@ def verify_gtp_no_echo(self, to_send):
...
+            remote_ip = self.sgi_remote_ip
+        to_send = self.IP(src=self.ue_ip, dst=remote_ip) / \
+            l4proto(sport=ue_port, dport=remote_port, **kwargs)
+        if payload is not None:
+            to_send /= Raw(payload)
+        return to_send
+
+    def assert_packet_sent_to_remote(self,  payload=None, l4proto=UDP,
+                                  ue_port=12345, remote_port=23456,
+                                  remote_ip=None):
```
ivan4th, line 1093: Doesn't this function duplicate `PFCPHelper.assert_packet_sent_to_sgi()`?

upf/test/test_upf.py
```diff
@@ -1048,6 +1061,81 @@ def verify_gtp_no_echo(self, to_send):
...
+        if payload is not None:
+            if isinstance(payload, bytes):
+                payload = Raw(payload)
+            to_send /= payload
+        self.if_sgi.add_stream(to_send)
+        self.pg_enable_capture(self.pg_interfaces)
+        self.pg_start()
+        return
+
+    def send_from_ue_to_sgi(self, payload=None, l4proto=UDP, ue_port=12345,
```
ivan4th, line 1081: This function doesn't appear to send anything. So, rename? (and maybe extract to a PFCPHelper method that should also be used as part of other `send_` functions)

upf/test/test_upf.py
```diff
@@ -1105,6 +1202,85 @@ class TestPGWIPv6(IPv6Mixin, TestPGWBase, framework.VppTestCase):
...
+    def ue_ip(self):
+        return self.UE_IP
+
+    @property
+    def remote_ip(self):
+        return APP_RULE_IP_V4
+
+    @property
+    def expected_length(self):
+        return 32
```
ivan4th, line 1228: Just use `expected_length = 32` at the top of the class

upf/test/test_upf.py
```diff
@@ -1105,6 +1202,85 @@ class TestPGWIPv6(IPv6Mixin, TestPGWBase, framework.VppTestCase):
     def ue_ip(self):
         return self.UE_IP
 
+    @property
+    def remote_ip(self):
+        return APP_RULE_IP_V6
+
+    @property
+    def expected_length(self):
+        return 52
```
ivan4th, line 1211: Just use `expected_length = 52` at the top of the class

upf/test/test_upf.py
```diff
@@ -1097,6 +1186,14 @@ class TestPGWIPv4(IPv4Mixin, TestPGWBase, framework.VppTestCase):
     def ue_ip(self):
         return self.UE_IP
 
+    @property
+    def remote_ip(self):
+        return APP_RULE_IP_V4
+
+    @property
+    def expected_length(self):
+        return 32
```
ivan4th, line 1195: Just use expected_length = 52 at the top of the class

upf/test/test_upf.py
```diff
@@ -946,6 +958,7 @@ class TestPGWBase(PFCPHelper):
 
     CLIENT_TEID = 1
     UNODE_TEID = 1127415596
+    teid = UNODE_TEID
```
ivan4th, line 961: Is this value modified somewhere in the test? If so, it must be set in `setUp()`, for example, or in the test methods.
Otherwise, it introduces dependency upon the ordering of the test run

ivan4th (changes requested):

upf/pfcp.c
```diff
@@ -781,8 +781,21 @@ format_f_teid (u8 * s, va_list * args)
...
+    s = format (s, "%d,V4:1,V6:1,CH:1", v->teid);
   else if ((v->flags & 0xf) == (F_TEID_CH | F_TEID_CHID))
     s = format (s, "%d,CH:1,CHID:%d", v->teid, v->choose_id);
+  else if ((v->flags & 0xf) == (F_TEID_CH | F_TEID_CHID | F_TEID_V4))
+    s = format (s, "%d,V4:1,CH:1,CHID:%d", v->teid, v->choose_id);
+  else if ((v->flags & 0xf) == (F_TEID_CH | F_TEID_CHID | F_TEID_V6))
+    s = format (s, "%d,V6:1,CH:1,CHID:%d", v->teid, v->choose_id);
+  else if ((v->flags & 0xf) == (F_TEID_CH | F_TEID_CHID | F_TEID_V4 |
+				F_TEID_V6))
+    s = format (s, "%d,V4:1,V6:1,CH:1,CHID:%d", v->teid, v->choose_id);
```
ivan4th, line 798: This part could be rewritten to check each flag in turn separately, adding corresponding parts like `V4:1` etc one by one

RoadRunnr: you could even do it in one go using the `!!` trick:

```
format (s, "%d,V4:%d,V6:%d",
    !!(v->flags & F_TEID_V4), !!(v->flags & F_TEID_V6),!!(v->flags & F_TEID_CH))
```

You could also first print all flags and then add TEID, CH-Id and the IPs to the end if the flags are set. That would be slightly more verbose than the old format, but would work.

sergeymatov: Refactored this debug function

upf/upf_pfcp_api.c
```diff
@@ -799,9 +805,163 @@ lookup_nwi (u8 * name)
...
+	  retry_cnt--;
+	  continue;
+	}
+
+      gtm->rand_base = teid;
+
+      ok =
+	((flags & F_TEID_V4) || (flags & F_TEID_V6)) &&
+	((teid_v4_lookup (gtm, teid, res) != 0) &&
+	 ((teid_v6_lookup (gtm, teid, res) != 0)));
```
RoadRunnr, line 869: I gave an example for this in https://github.com/travelping/upg-vpp/pull/3#discussion_r506163343.

What made you think that your version is correct or semantically equivalent to my suggestion?

```suggestion
      ok =
	(!(flags & F_TEID_V4) || teid_v4_lookup (gtm, teid, res) == 0) &&
	(!(flags & F_TEID_V6) || teid_v6_lookup (gtm, teid, res) == 0);
```

sergeymatov: Way I've done is kind of a an obvious expression of desired logic
OK = false if teid lookup is failed for any of given teid+endpoint ip combination regardless it's version. And true otherwise.

If you will take a look at teid_v4/6_lookup function, it returns UPF_GTPU_ERROR_NO_SUCH_TUNNEL if there is no tunnel with given teid/res, 0 if there is one

I was having doubts about logic you've introduced. Let's will pick up simplest scenario with V4 flag and there is no tunnel created for given teid, and for simplified pipeline let's call (flags & F_TEID_V4) as 1(basically every int not equal to 0), and UPF_GTPU_ERROR_NO_SUCH_TUNNEL as -1 without breaking general logic
so we have
(!1 || -1 == 0) && (!0 || -1 ==0)
which as a result gives us 0 for any of teid generated. However, we do expecting 1 here cause V4 flag is given and no tunnel is yet created for given combination of teid and endpoint ip

sergeymatov: Logic you've introduced is fine if `==` change to `!=`, probably you've just missed return value of teid lookup.
I've changed this and left a comment, that teid lookup for both ip4/ip6 is safe even there is no IP4/IP6 address configured for upip_res.

RoadRunnr: I don't know why you need that comment. The lookup will only be invoked if corresponding flag is set.

But it doesn't hurt to have it, so no need to remove it.

RoadRunnr (changes requested):
there seems to be multiple whitespace problems (mostly tab vs. spaces), not sure how this passed through the formater check

upf/pfcp.c
```diff
@@ -794,61 +793,65 @@ decode_f_teid (u8 * data, u16 length, void *p)
...
+      if (!(v->flags & F_TEID_V6) && !(v->flags & F_TEID_V4))
+	{
+	  pfcp_debug
+	    ("PFCP: F-TEID with CH flag should have at least v4/v6 flag:%02x",
+	     v->flags);
+	  return -1;
+	}
+
       v->choose_id = get_u8 (data);
     }
```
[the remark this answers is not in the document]
sergeymatov: By definition it's expected CH flag set only in CP function. So far can't imagine case when it's should be handled in upg

upf/upf_pfcp.c
```diff
@@ -72,9 +73,12 @@ format_upf_device_name (u8 * s, va_list * args)
   u32 i = va_arg (*args, u32);
   upf_nwi_t *nwi;
 
-  nwi = pool_elt_at_index (gtm->nwis, i);
+  if (!pool_is_free_index (gtm->nwis, i))
+    {
+      nwi = pool_elt_at_index (gtm->nwis, i);
+      s = format (s, "upf-nwi-%U", format_network_instance, nwi->name);
+    }
```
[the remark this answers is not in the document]
sergeymatov: This can be happen if we will delete nwi with configured UPIP and then try device output CLI command
as example - "show hardware"
I will move fixes of nwi deletion in a separate PR

upf/pfcp.c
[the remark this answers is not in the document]
sergeymatov: Probably there is a difference in how it looks in PR and original view.
Checkstyle works fine for this file

upf/upf_pfcp_api.c
```diff
@@ -799,9 +805,166 @@ lookup_nwi (u8 * name)
...
+static u32
+process_teid_generation (upf_main_t * gtm, u8 chid, u32 flags,
+			 upf_upip_res_t * res, upf_session_t * sx)
+{
+  u8 retry_cnt = 10;
+  u32 teid = 0;
+  bool ok = false;
+  do
+    {
+      teid = random_u32 (&gtm->rand_base) & (UINT32_MAX ^ res->mask);
```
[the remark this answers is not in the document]
sergeymatov: yep, this has same result
fixed

upf/upf_cli.c
```diff
@@ -582,6 +582,39 @@ VLIB_CLI_COMMAND (upf_tdf_ul_enable_command, static) = {
...
+      if (unformat (line_input, "release %u", &spec_version))
+	break;
+      else
+	return 0;
+    }
+
+  gtm->pfcp_spec_version = spec_version;
+
+  return NULL;
+}
```
RoadRunnr, line 608: a show command wouldn't be too bad.

upf/upf_pfcp.c
```diff
@@ -541,6 +542,10 @@ pfcp_create_session (upf_node_assoc_t * assoc,
   node_assoc_attach_session (assoc, sx);
   hash_set (gtm->session_by_id, cp_seid, sx - gtm->sessions);
 
+  /*Init TEID by choose_id hash lookup table */
+  sx->teid_by_chid =
```
RoadRunnr, line 546: the CLI debug command `show upf session` should also dump this

RoadRunnr (changes requested):
overall, LGTM, two smallish additions.

Tests with FTUP enabled ergw look good so far.

upf/upf_pfcp.c
```diff
@@ -2630,6 +2630,20 @@ format_pfcp_session (u8 * s, va_list * args)
 	      vlib_time_now (gtm->vlib_main) - sx->last_ul_traffic,
 	      rules->inactivity_timer.handle);
 
+  if (gtm->pfcp_spec_version == 16)
+    {
+      s = format (s, "TEID assignment per choose ID\n");
+      u16 idx = 1;
+      do
```
RoadRunnr, line 2637: a plain `for` loop would be simpler

upf/upf_pfcp.c
```diff
@@ -2630,6 +2630,20 @@ format_pfcp_session (u8 * s, va_list * args)
...
+	{
+	  u32 teid = *sparse_vec_elt_at_index (sx->teid_by_chid, idx);
+	  if (teid)
+	    s = format (s, "Choose ID [%u] TEID [%u]\n", idx, teid);
```
RoadRunnr, line 2641: 1.) indent that properly, 2 spaces more than the header in line 2635
2.) its a bit verbose maybe:
```
s = format (s, "    %3u: %u (0x%08x)\n", idx, teid, teid);
```

sergeymatov: Sorry, completely forgot about output indentation
fixed

ivan4th:
@RoadRunnr so all of the objections fixed atm? Can we merge this PR?

RoadRunnr:
> @RoadRunnr so all of the objections fixed atm? Can we merge this PR?

C parts are fine. If you are ok with the tests, then it can be merged.

RoadRunnr (approved):
C parts look good to me

PR #18, opened 2020-10-26
UPG counters

sergeymatov:
This commit adds small set of UPG-related counters.
Based on vlib_counter lib.

### DIFF
--- a/upf/flowtable.c
+++ b/upf/flowtable.c
@@ -168,11 +168,17 @@ expire_single_flow (flowtable_main_t * fm, flowtable_main_per_cpu_t * fmt,
     }
   else
     {
+      upf_main_t *gtm = &upf_main;
       upf_debug ("Flow Remove %p", f);
       pool_put (fmt->timers, e);
 
       /* hashtable unlink */
       flowtable_entry_remove (fmt, f);
+
+      vlib_decrement_simple_counter (&gtm->upf_simple_counters
+				     [UPF_FLOW_COUNTER],
+				     vlib_get_thread_index() ,
+				     fmt - fm->per_cpu, 1);
 
       /* free to flow cache && pool (last) */
       flow_entry_free (fm, fmt, f);
@@ -272,6 +278,7 @@ flowtable_entry_lookup_create (flowtable_main_t * fm,
 {
   flow_entry_t *f;
   dlist_elt_t *timer_entry;
+  upf_main_t *gtm = &upf_main;
 
   if (PREDICT_FALSE
       (BV (clib_bihash_search_inline) (&fmt->flows_ht, kv) == 0))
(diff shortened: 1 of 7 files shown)


ivan4th:
It should be rather easy to verify the counters from test_upf.py, so probably that should be done

upf/flowtable_init.c
```diff
@@ -80,6 +83,14 @@ flowtable_init_cpu (flowtable_main_t * fm, flowtable_main_per_cpu_t * fmt)
...
   upf_debug ("POOL SIZE %u", pool_elts (fmt->timers));
 
+  /* Init flows counter per cpu */
+  upf_counter_lock (gtm);
+  vlib_validate_simple_counter (&gtm->upf_simple_counters[UPF_FLOW_COUNTER],
+				fmt - fm->per_cpu);
+  vlib_zero_simple_counter (&gtm->upf_simple_counters[UPF_FLOW_COUNTER],
+			    fmt - fm->per_cpu);
+  upf_counter_unlock (gtm);
+
```
RoadRunnr, line 93: putting the counters in flowtable_main_t might make more sense. simple_counters are per CPU thread, repeating that by using the index and CPU id does not make sense.

Or is there a specific reason you want to read the per CPU counters?

upf/upf.c
```diff
@@ -359,6 +359,29 @@ upf_init (vlib_main_t * vm)
 					FIB_SOURCE_PRIORITY_HI,
 					FIB_SOURCE_BH_SIMPLE);
 
+  clib_spinlock_init (&sm->upf_counter_lock);
```
RoadRunnr, line 362: the look is not really needed.
Simple counters are per CPU anyway and if you don't use the index as per CPU id, then no locking is required.

RoadRunnr (changes requested):

upf/flowtable_init.c
sergeymatov: I've just following your suggestion to make it per-cpu, but I suppose it was only suitable for gauges.
Keeping in mind that counters are per-cpu vectored more logical is to have in in flowtable_main for sure.
Will be changed.

upf/upf.c
sergeymatov: sure. will be fixed

upf/flowtable.c
```diff
@@ -168,12 +168,17 @@ expire_single_flow (flowtable_main_t * fm, flowtable_main_per_cpu_t * fmt,
...
+      upf_main_t *gtm = &upf_main;
       upf_debug ("Flow Remove %p", f);
       pool_put (fmt->timers, e);
 
       /* hashtable unlink */
       flowtable_entry_remove (fmt, f);
 
+      vlib_decrement_simple_counter (&gtm->upf_simple_counters
+				     [UPF_FLOW_COUNTER],
+				     0, fmt - fm->per_cpu, 1);
```
RoadRunnr, line 180: this and the other places where the FLOW_COUNTER is used are wrong. You mixed up the `thread_index` with the `index`

```suggestion
      vlib_decrement_simple_counter (&gtm->upf_simple_counters
				     [UPF_FLOW_COUNTER],
				     fmt - fm->per_cpu, 0, 1);
```

@ivan4th what do think about the way that is used to calculate the `thread_index`? `fmt - fm->per_cpu` is correct, but also quite strange.
I would rather prefer to use `vlib_get_thread_index ()` instead.

ivan4th: I agree that `vlib_get_thread_index ()` is better there

ivan4th:
@RoadRunnr all objections fixed as expected?

RoadRunnr (approved):
LGTM

PR #19, opened 2020-10-26
Bump indent to 2.2.12

sergeymatov:


Dockerfile.build
```diff
@@ -20,6 +20,11 @@ RUN --mount=target=/var/lib/apt/lists,type=cache,sharing=private \
     tar -xvz -C /usr/local bin/buildctl && \
     echo "${BUILDCTL_SHA256}  /usr/local/bin/buildctl" | sha256sum -c && \
     cd /src/vpp && \
+    apt-get install  -y --no-install-recommends wget && \
```
ivan4th, line 23: we have curl above already, why not use
`curl -sSL -O http://mirrors.kernel.org/ubuntu/pool/main/i/indent/indent_2.2.12-1_amd64.deb` ?

also, it would be nice to do SHA256 check for the downloaded `deb` like it's done for `buildctl` binary above

sergeymatov: added sha check, removed wget installation/usage

ivan4th:
LGTM

PR #22, opened 2020-11-03
Implement IP-based ADF rules

sergeymatov:
This is cherry-pick of commit 2301de5184af8810c862481ef403562de90ed9bd
authored by @ivan4th

### DIFF
--- a/upf/CMakeLists.txt
+++ b/upf/CMakeLists.txt
@@ -48,6 +48,7 @@ if(HS_FOUND)
     upf_tcp_forward.c
     upf_proxy.c
     upf_app_db.c
+    upf_ipfilter.c
     flowtable_init.c
     flowtable.c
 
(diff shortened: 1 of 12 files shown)


ivan4th:
Previous discussion https://github.com/travelping/vpp/pull/37
IIRC it was good to go except for merge conflicts which are resolved in this PR.
@RoadRunnr can we merge this PR?

RoadRunnr:
> @RoadRunnr can we merge this PR?

I'll leave that to you. My two concerns (that need not block a merge):

* binary API support, do we need that?
* the rule/ip DB support feels not good, but nothing I can put my finger on.

ivan4th:
My take on that is that basically this can be done as followups.
Binary API we do need, so there will be PR for it.
As of DB support, there's possibility that this part will have to be updated when we do another required followup, namely, adding a possibility to update the rules "on the fly", when there are active sessions for the rules that are being modified.

PR #23, opened 2020-11-11
Add new UPG E2E tests

ivan4th:
This PR adds new Go-based UPG E2E tests, replacing old external test usage.

The tests are based upon the idea of creating several network namespaces and then using UPG to pass traffic between them.

The test framework includes:
* "Captive VPP" code for running VPP as a subprocess in a separate network namespace
* PFCP state machine
* Traffic generator (HTTP / UDP ping / ICMP ping)
* GTP-U tunnel support (with kernel mode in case of IPv4)
* Network namespace utils

Rationale for using Go for E2E:
* Go strikes the balance between speed and ease of maintenance, as from my experience Python is too slow for timing-dependent tests, writing all of the tests in C would take too much effort
* the test framework can be reused by k8s-related projects written in Go that make use of UPG
* there are Go libraries for packet capture/manipulation, PFCP and GTP

Rationale for inclusion of E2E in UPG source tree as opposed to using a separate repo:
* the tests are intended to evolve with UPG, with new test cases being added for each bugfix and feature
* Python-based VPP integration tests are not very well-suited for writing UPG tests as they need rather advanced test harness to function, and also the test cases are often very timing-sensitive
* given that most of the tests (excluding unit tests) have to be written in other language anyway, having Go tests in the repository might not be that bad

The test framework can possibly be extracted to a separate repository once
it stabilizes, as it can be used by other projects that make use of UPG.

TODO:
- [X] Add the tests
- [x] Make gitlab tests pass
- [x] Remove E2E_FOCUS on gitlab
- [x] Update README.md


hack/buildenv.sh
```diff
@@ -6,20 +6,36 @@ set -o errtrace
...
+         ${priv} \
          -v $PWD:/src:delegated \
          -v $PWD/vpp-out:/vpp-out \
-         -e LC_ALL=C.UTF-8 -e LANG=C.UTF-8 \
+         -e LC_ALL=C.UTF-8 \
+         -e LANG=C.UTF-8 \
+         -e E2E_RETEST="${E2E_RETEST:=}" \
+         -e E2E_PARALLEL="${E2E_PARALLEL:-}" \
+         -e E2E_PARALLEL_NODES="${E2E_PARALLEL_NODES:-}" \
+         -e E2E_FOCUS="${E2E_FOCUS:-}" \
```
sergeymatov, line 33: Can we have all possible frags of regex listed in README for tests?

sergeymatov (approved):
LGTM

PR #33, opened 2020-11-25
Fix/pfcp memleak

sergeymatov:


RoadRunnr (approved):
LGTM

ivan4th:
LGTM, too.

upf/upf_pfcp_api.c
```diff
@@ -2852,6 +2854,7 @@ handle_session_modification_request (pfcp_msg_t * req,
   upf_pfcp_send_response (req, cp_seid, PFCP_SESSION_MODIFICATION_RESPONSE,
 			  &resp.grp);
 
+  vec_free (resp.created_pdr);
```
RoadRunnr, line 2857: that can't be right, upf_pfcp_send_response is calling pfcp_free_msg and that should actually free that.

Something else must be broken if created_pdr is not freed.

RoadRunnr (changes requested):

PR #34, opened 2020-11-25
Only set CREATED_PDR group if FTUP was enabled

sergeymatov:
Also revert freeing of created_pdr vec for response on session
modification/establishment request.

### DIFF
--- a/upf/upf_pfcp_api.c
+++ b/upf/upf_pfcp_api.c
@@ -677,7 +677,7 @@ process_teid_generation (upf_main_t * gtm, u8 chid, u32 flags,
 static int
 handle_f_teid (upf_session_t * sx, upf_main_t * gtm, pfcp_pdi_t * pdi,
 	       upf_pdr_t * process_pdr, pfcp_created_pdr_t ** created_pdr_vec,
-	       upf_upip_res_t * res, struct pfcp_group *grp, u8 create)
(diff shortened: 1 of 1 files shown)


ivan4th:
LGTM

upf/upf_pfcp_api.c
```diff
@@ -728,7 +728,6 @@ handle_f_teid (upf_session_t * sx, upf_main_t * gtm, pfcp_pdi_t * pdi,
 	{
 	  vec_add2 (*created_pdr_vec, created_pdr, 1);
 	  memset (created_pdr, 0, sizeof (*created_pdr));
-	  SET_BIT (grp->fields, SESSION_ESTABLISHMENT_RESPONSE_CREATED_PDR);
```
RoadRunnr, line 731: This removes the only user of the `grp` function argument. The argument should also be removed.

RoadRunnr (changes requested):
While reviewing this change, I also discovered that handling of the `create` argument to handle_f_teid is broken. Will open a seperate issue.

upf/upf_pfcp_api.c
sergeymatov: fixed, thanks for note

RoadRunnr (approved):
LGTM

ivan4th:
LGTM again :)

PR #36, opened 2020-12-01
Don't handle F-TEID on Update PDR

sergeymatov:
For Update PDR we are not doing F-TEID processing. If
PDI does contain matchinf for F-TEID to be changed we just
apply IE as is.

### DIFF
--- a/upf/upf_pfcp_api.c
+++ b/upf/upf_pfcp_api.c
@@ -771,7 +771,6 @@ handle_create_pdr (upf_session_t * sx, pfcp_create_pdr_t * create_pdr,
 
   rules = pfcp_get_rules (sx, PFCP_PENDING);
   vec_alloc (rules->pdr, vec_len (create_pdr));
-  vec_alloc (*created_pdr_vec, vec_len (create_pdr));
 
   vec_foreach (pdr, create_pdr)
   {
--- a/upf/upf_pfcp_api.c
+++ b/upf/upf_pfcp_api.c
@@ -702,6 +702,9 @@ handle_f_teid (upf_session_t * sx, upf_main_t * gtm, pfcp_pdi_t * pdi,
 
       if (!chosen)
 	{
+	  if (!create)
+	    return -1;
+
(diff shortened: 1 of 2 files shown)


upf/upf_pfcp_api.c
```diff
@@ -1025,12 +1023,7 @@ handle_update_pdr (upf_session_t * sx, pfcp_update_pdr_t * update_pdr,
 
     if (ISSET_BIT (pdr->pdi.grp.fields, PDI_F_TEID))
       {
-	if (handle_f_teid
-	    (sx, gtm, &pdr->pdi, update, NULL, res, grp, 0) != 0)
-	  {
-	    r = -1;
-	    break;
-	  }
+	update->pdi.teid = pdr->pdi.f_teid;
```
RoadRunnr, line 1026: no, the `F-TEID` in a `Update PDR` can contain a reference to a CHID. That CHID must have been use in a `Create PDR` before and reference a valid TEID. 
The lookup part in handle_f_teid must still be executed, but should fail if doesn't find an existing TEID.

RoadRunnr (changes requested):

upf/upf_pfcp_api.c
```diff
@@ -698,6 +698,9 @@ handle_f_teid (upf_session_t * sx, upf_main_t * gtm, pfcp_pdi_t * pdi,
 
 	  if (teid)
 	    chosen = true;
+
+	  if (!(chosen) && !(create))
+	    return -1;
```
RoadRunnr, line 703: that should IMHO be a few lines down, after the end of the `if ((pdi->f_teid.flags & F_TEID_CHID))`

A F-TEID in `Update-PDR` with F_TEID_CH = 1, F_TEID_CHID = 0 is IMHO illegal. By moving the chosen+create test after the F_TEID_CH check, a proper error for that combination is ensured.

RoadRunnr (changes requested):

RoadRunnr (approved):
LGTM

PR #37, opened 2020-12-02
Handle duplicate UE IP+TEIDs across multiple sessions

### LINKED ISSUE #25: Crash on duplicate UE IP
UPG crashes if there's duplicate UE IP between several sessions. Most likely, this only concerns duplicate UE IP Address between several sessions on PDIs for SGi an also possibly for Access in TDF mode, but duplicate UE IPs in PGW mode must also be checked.

ivan4th:
This patch makes sure that a PDIs don't conflict for the PDIs for the
same NWI from other sessions based on UE IPs and TEIDs. This is not
yet a complete implementation of the spec in this respect b/c other
IEs in PDI also need to be taken into account.

Previously, adding many duplicate UE IPs could cause UPG to
crash (about 255) due to FIB entry src count limit, and, in addition
to that, deleting sessions with duplicate UEs could lead to an
assertion failure due to a bad session reference.

Now, if a conflict is detected, the corresponding Session Establishment Request
or Session Modification Request fails with "rule creation/modification failure" as the cause.

Fixes #25.

- [X] detect duplicate UE IP+TEIDs across multiple sessions
- [X] E2E test for TDF case (no TEIDs, just UE IPs)
- [x] E2E test for GTP-U case
- [x] verify session modification
- [x] get rid of unrelated indent changes

### DIFF
--- a/upf/upf_pfcp_api.c
+++ b/upf/upf_pfcp_api.c
@@ -596,38 +596,40 @@ validate_teid (u32 teid)
   return ((teid != 0) && (teid != 0xffffffff));
 }
 
-static int
-teid_v4_lookup (upf_main_t * gtm, u32 teid, upf_upip_res_t * res)
+static u32
+teid_v4_lookup_session_index (upf_main_t * gtm, u32 teid, ip4_address_t * ip4)
 {
   clib_bihash_kv_8_8_t kv, value;
   gtpu4_tunnel_key_t key4;
 
-  key4.dst = res->ip4.as_u32;
+  key4.dst = ip4->as_u32;
   key4.teid = teid;
 
   kv.key = key4.as_u64;
 
   if (PREDICT_FALSE
       (clib_bihash_search_8_8 (&gtm->v4_tunnel_by_key, &kv, &value)))
-    return UPF_GTPU_ERROR_NO_SUCH_TUNNEL;
+    return ~0;
 
-  return 0;
+  /* only lower 32 bits are used here */
+  return value.value;
 }
 
-static int
-teid_v6_lookup (upf_main_t * gtm, u32 teid, upf_upip_res_t * res)
+static u32
+teid_v6_lookup_session_index (upf_main_t * gtm, u32 teid, ip6_address_t * ip6)
 {
   clib_bihash_kv_24_8_t kv, value;
 
-  kv.key[0] = res->ip6.as_u64[0];
-  kv.key[1] = res->ip6.as_u64[1];
+  kv.key[0] = ip6->as_u64[0];
+  kv.key[1] = ip6->as_u64[1];
   kv.key[2] = teid;
 
   if (PREDICT_FALSE
       (clib_bihash_search_24_8 (&gtm->v6_tunnel_by_key, &kv, &value)))
-    return UPF_GTPU_ERROR_NO_SUCH_TUNNEL;
+    return ~0;
 
-  return 0;
+  /* only lower 32 bits are used here */
+  return value.value;
 }
 
 static u32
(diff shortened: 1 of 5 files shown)


RoadRunnr:
I wonder if the UE IP check really needs a hash. Wouldn't it suffice to perform a FIB lookup on the SGi/N6 side fib table and reject the request if we find a /32 (/64 IPv6) entry for the IP that points to a different session?

ivan4th:
@RoadRunnr I thought about that too initially, but note that this PR also checks for duplicate UE IP+TEID pairs on NWIs, in which case there are no session DPOs.

RoadRunnr:
> @RoadRunnr I thought about that too initially, but note that this PR also checks for duplicate UE IP+TEID pairs on NWIs, in which case there are no session DPOs.

Yes, but a duplicate TEID + UE IP within the same session are valid. I'm not sure what exactly you want to prevent with that check?

A duplicate Fq-TEID (GTP-IP + TEID) across multiple sessions is not valid regardless of the UE IP, but that should be handled already.

ivan4th:
Hmm, probably I've missed TEID duplication check across multiple sessions. `handle_f_teid()` doesn't appear to have one https://github.com/travelping/upg-vpp/blob/master/upf/upf_pfcp_api.c#L677-L750 nor do `handle_create_pdr()` / `handle_update_pdr()`. Will check a bit more. The idea is to prevent the duplication early, before applying any changes to the session. Although, using UE IP for that is incorrect indeed, `v[46]_tunnel_by_key` should be used instead

RoadRunnr:
> Hmm, probably I've missed TEID duplication check across multiple sessions. `handle_f_teid()` doesn't appear to have one https://github.com/travelping/upg-vpp/blob/master/upf/upf_pfcp_api.c#L677-L750 nor do `handle_create_pdr()` / `handle_update_pdr()`. Will check a bit more. The idea is to prevent the duplication early, before applying any changes to the session. Although, using UE IP for that is incorrect indeed, `v[46]_tunnel_by_key` should be used instead

You are right, there is not check. I thought there was one. Duplicate TEID are prevented to create havoc by the the fact that the {v4,v6}_tunnel_by_key will not accept duplicate keys, newer tunnel will always replace existing tunnels.

We should probably check all use case of bihashes, change the is_add argument to 2 (meaning add but not overwrite) and check the return.

Checking the hashes for collisions is a good idea. Putting the check into handle_f_teid() https://github.com/travelping/upg-vpp/blob/master/upf/upf_pfcp_api.c#L745 looks like the best place. The TEID generation can not generate colliding TEIDs by design.

upf/upf_pfcp_api.c
```diff
@@ -596,38 +596,38 @@ validate_teid (u32 teid)
...
 
   kv.key = key4.as_u64;
 
   if (PREDICT_FALSE
       (clib_bihash_search_8_8 (&gtm->v4_tunnel_by_key, &kv, &value)))
-    return UPF_GTPU_ERROR_NO_SUCH_TUNNEL;
+    return ~0;
 
-  return 0;
+  return value.value & 0xffffffff;
```
RoadRunnr, line 614: what's that for? The u64 should be truncated automatically to u32 on return.

upf/upf_pfcp_api.c
```diff
@@ -736,14 +739,38 @@ handle_f_teid (upf_session_t * sx, upf_main_t * gtm, pfcp_pdi_t * pdi,
...
   return 0;
 
 }
 
+static int
+validate_ue_ip (upf_session_t * sx, upf_nwi_t * nwi,
+		pfcp_ue_ip_address_t * ue_addr)
+{
+  upf_main_t *gtm = &upf_main;
+  const dpo_id_t *dpo = ue_addr->flags & IE_UE_IP_ADDRESS_V4 ?
```
RoadRunnr, line 767: Sadly, UE IP can contain both IPv4 *and* IPv6. There is no restriction in TS 29.244 on the combination of the flags.

The only thing that could interpreted as a restriction is this:

>  It contains a source or destination IP address

Its in singular and that would im indicate the two addresses are not expected.

A quick search shows that the other places that deal with UE IP Address IEs except instances with both IP versions set.

ivan4th: Fixed, thanks

upf/upf_pfcp_api.c
ivan4th: Removed that `&` part

RoadRunnr (approved):
LGTM

PR #38, opened 2020-12-02
add node id support

5G-Center-Chinatelecom:


sergeymatov:
Please check Pull Request #52 with more detailed implementation
If no objections from committer side I would like to close this PR and keep activities in PR #52

RoadRunnr:
alternate version with CLI and FQDN in Node-Id support in PR #52, please follow up on that if you have comments.

PR #40, opened 2020-12-08
Fix TEID lookup per choose_id

sergeymatov:


upf/upf_pfcp_api.c
```diff
@@ -694,7 +694,7 @@ handle_f_teid (upf_session_t * sx, upf_main_t * gtm, pfcp_pdi_t * pdi,
       if ((pdi->f_teid.flags & F_TEID_CHID))
 	{
 	  u8 chid = pdi->f_teid.choose_id;
-	  teid = sparse_vec_index (sx->teid_by_chid, chid);
+	  teid = *sparse_vec_elt_at_index (sx->teid_by_chid, chid);
 
 	  if (teid)
 	    chosen = true;
```
RoadRunnr, line 700: * chosen is not required, teid can be used directly since choosing TEID = 0 is not valid
* it might be slightly clearer to use sparse_vec_index and test for the zero index instead of relying on the index zero element being 0

```suggestion
      if ((pdi->f_teid.flags & F_TEID_CHID))
	{
	  uword i = sparse_vec_index (sx->teid_by_chid, pdi->f_teid.choose_id);

	  if (i)
	    teid = vec_elt (sx->teid_by_chid, i);
	}
```

sergeymatov: Thanks, done

RoadRunnr (approved):
LGTM for C part, can't comment on unit test.

upf/test/test_upf.py
```diff
@@ -1182,12 +1185,17 @@ def remote_ip(self):
...
     UE_IP = "198.20.0.2"
+    UE_IP2 = "10.10.0.2"
     expected_length = 32
 
     @property
     def ue_ip(self):
         return self.UE_IP
 
+    @property
+    def ue_ip2(self):
```
ivan4th, line 1196: this property is not needed, see below

upf/test/test_upf.py
```diff
@@ -1265,6 +1274,26 @@ def forward_pdr(self, pdr_id=1):
             IE_URR_Id(id=pdr_id),
         ])
 
+    def forward_pdr2(self, pdr_id=1):
```
ivan4th, line 1277: It's probably better to have
`def forward_pdr(self, pdr_id=1, chid=201, ue_ip=None):`

`def ue_ip2(self):` should probably be removed, the reason for `def ue_ip(self):` is a separate impl for IPv6, and this is a concrete IPv4 test class. Just use `forward_pdr(pdr_id=3, chid=201, self.UE_IP2)` instead of `forward_pdr2(pdr_id=3)`

upf/test/test_upf.py
```diff
@@ -1220,6 +1228,7 @@ def establish_session(self):
             ]),
             self.forward_pdr(pdr_id=1),
             self.forward_pdr(pdr_id=2),
+            self.forward_pdr2(pdr_id=3),
```
ivan4th, line 1231: `forward_pdr(pdr_id=3, chid=201, UE_IP2)`, see below

upf/test/test_upf.py
```diff
@@ -147,6 +147,9 @@ def ie_fteid_ch(self, chid):
     def ie_ue_ip_address(self, SD=0):
         return IE_UE_IP_Address(ipv4=self.ue_ip, V4=1, SD=SD)
 
+    def ie_ue_ip_address2(self, SD=0):
```
ivan4th, line 150: Remove this func, just add an optional `ip=None` argument to `ie_ue_ip_address()`

upf/test/test_upf.py
```diff
@@ -147,6 +147,9 @@ def ie_fteid_ch(self, chid):
     def ie_ue_ip_address(self, SD=0):
         return IE_UE_IP_Address(ipv4=self.ue_ip, V4=1, SD=SD)
```
ivan4th, line 148:
```python
    def ie_ue_ip_address(self, SD=0, ip=None):
         return IE_UE_IP_Address(ipv4=ip or self.ue_ip, V4=1, SD=SD)
```
(and remove `ie_ue_ip_address2()`, see below)

upf/test/test_upf.py
ivan4th: `self.ue_ip_ip_address()` should be `self.ue_ip_ip_address(ue_ip=ue_ip)` in this case

ivan4th (changes requested):

ivan4th:
LGTM.

From discussion elsewhere: the Python tests were not intended to be committed, and this part (as well as rest of the FTUP feature) is to be tested using E2E later.

ivan4th (approved):

ivan4th:
Pushed to gitlab by hand to run the tests ... As GitLab mirroring, and thus GitLab CI integration appears to be broken

Issue #44, opened 2020-12-17
Crash in response_expired()


ivan4th:
The following crash has been observed when running e2e tests:
```
INFO[0827] VPP: /src/vpp/src/plugins/upf/upf_pfcp_server.h:240 (pfcp_msg_pool_elt_at_index) assertion `m->is_valid_pool_item' fails  pid=5954 what=stdout
INFO[0827] VPP:                                          pid=5954 what=stdout
INFO[0827] VPP: Program received signal SIGABRT, Aborted.  pid=5954 what=stdout
INFO[0827] VPP: 0x00007ffff4717fb7 in raise () from /lib/x86_64-linux-gnu/libc.so.6  pid=5954 what=stdout
INFO[0827] VPP: #0  0x00007ffff4717fb7 in raise () from /lib/x86_64-linux-gnu/libc.so.6  pid=5954 what=stdout
INFO[0827] VPP: #1  0x00007ffff4719921 in abort () from /lib/x86_64-linux-gnu/libc.so.6  pid=5954 what=stdout
INFO[0827] VPP: #2  0x0000000000407193 in os_panic () at /src/vpp/src/vpp/vnet/main.c:371  pid=5954 what=stdout
INFO[0827] VPP: #3  0x00007ffff55f8619 in debugger () at /src/vpp/src/vppinfra/error.c:84  pid=5954 what=stdout
INFO[0827] VPP: #4  0x00007ffff55f8397 in _clib_error (how_to_die=2, function_name=0x0, line_number=0, fmt=0x7fffac0ba8b7 "%s:%d (%s) assertion `%s' fails") at /src/vpp/src/vppinfra/error.c:143  pid=5954 what=stdout
INFO[0827] VPP: #5  0x00007fffac063138 in pfcp_msg_pool_elt_at_index (psm=0x7fffac2f7250 <pfcp_server_main>, index=11880) at /src/vpp/src/plugins/upf/upf_pfcp_server.h:240  pid=5954 what=stdout
INFO[0827] VPP: #6  0x00007fffac062052 in response_expired (id=11880) at /src/vpp/src/plugins/upf/upf_pfcp_server.c:462  pid=5954 what=stdout
INFO[0827] VPP: #7  0x00007fffac05352a in pfcp_process (vm=0x7ffff6483b80 <vlib_global_main>, rt=0x7fffb4cb3580, f=0x0) at /src/vpp/src/plugins/upf/upf_pfcp_server.c:1250  pid=5954 what=stdout
[...]

PR #45, opened 2020-12-17
Improve stability of E2E tests

ivan4th:
This PR includes the following e2e fixes:
* VPP ARP cache is pre-populated, reducing flakes in ICMP / UDP tests
* added proper timeouts for ICMP tests so they don't hang forever in case of failure
* made it possible to create much larger number of sessions at once, and updated the leak test to create 10k sessions at once
* stop skipping any tests on CI
* disabled flake retries on CI

Improving the leak test has uncovered a new issue (#44), so the CI still may fail intermittently on that test, but that is to be fixed shortly.

### DIFF
--- a/test/e2e/upg_e2e.go
+++ b/test/e2e/upg_e2e.go
@@ -369,13 +369,13 @@ var _ = ginkgo.Describe("Multiple PFCP Sessions", func() {
 			_, err := f.VPP.Ctl("memory-trace main-heap on")
 			framework.ExpectNoError(err)
 			var ueIPs []net.IP
-			for i := 0; i < 100; i++ {
+			for i := 0; i < 10000; i++ {
 				ueIPs = append(ueIPs, f.AddUEIP())
 			}
-			for i := 0; i < 100; i++ {
-				ginkgo.By("creating 100 sessions")
+			for i := 0; i < 3; i++ {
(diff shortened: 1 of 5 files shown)


test/e2e/upg_e2e.go
```diff
@@ -386,14 +386,14 @@ var _ = ginkgo.Describe("Multiple PFCP Sessions", func() {
...
+				ginkgo.By("deleting 10000 sessions")
 				for _, seid := range seids {
 					deleteSession(f, seid, false)
 				}
 			}
 
-			ginkgo.By("Waiting 10 seconds for the queues to be emptied")
-			time.Sleep(10 * time.Second)
+			ginkgo.By("Waiting 40 seconds for the queues to be emptied")
+			time.Sleep(40 * time.Second)
```
sergeymatov, line 396: 40 seconds timeout will surely work even on weak performing envs?

sergeymatov (approved):
LGTM

test/e2e/upg_e2e.go
ivan4th: When UPG receives a PFCP request, it caches the response for 30s. If the request is repeated (same seq), the cached response is sent again. After 30s, the entries from the cache are gone, and that is important for checking the leaks.

PR #48, opened 2020-12-22
e2e: add test for the GTP Proxy case

ivan4th:
Verify GTP-U <-> GTP-U case in addition to existing TDF and PGW test cases.
Make sure any GTP-U extensions are passed as-is, too.

In addition to that, several handy env vars are added for e2e tests (bundled in this PR as there's no such thing as PR chaining):
* `E2E_DISPATCH_TRACE` - store VPP dispatch trace in the test artifacts
* `E2E_TRACE` - enable VPP packet trace during test execution
* `E2E_PAUSE_ON_ERROR` - pause tests upon error, making it possible to debug them interactively


sergeymatov (approved):
Looks good to me, approved

PR #49, opened 2020-12-24
Fix deleting the PFCP responses for a session

### LINKED ISSUE #44: Crash in response_expired()

ivan4th:
**NOTE:** Before merging this PR, #48 must be merged, after which this PR needs to be carefully rebased by hand.

A response expiration timer firing at the same time as a session is
being deleted was causing UPG to crash.

This PR also adds an E2E test case that verifies the fix.

Fixes #44.

### DIFF
--- a/upf/upf_pfcp_server.c
+++ b/upf/upf_pfcp_server.c
@@ -387,6 +387,7 @@ static void
 request_t1_expired (u32 seq_no)
 {
   pfcp_server_main_t *psm = &pfcp_server_main;
+  upf_main_t *gtm = &upf_main;
   pfcp_msg_t *msg;
   uword *p;
 
@@ -417,8 +418,12 @@ request_t1_expired (u32 seq_no)
       hash_unset (psm->request_q, msg->seq_no);
       pfcp_msg_pool_put (psm, msg);
 
-      if (type == PFCP_HEARTBEAT_REQUEST)
-	upf_pfcp_server_deferred_release_association (msg->node);
+      if (type == PFCP_HEARTBEAT_REQUEST 
+ 	  && !pool_is_free_index (gtm->nodes, msg->node))
+	{
+	  upf_node_assoc_t *n = pool_elt_at_index (gtm->nodes, msg->node);
+	  pfcp_release_association (n);
+	}
     }
 }
 
@@ -1052,19 +1057,16 @@ u32 upf_pfcp_server_start_timer (u8 type, u32 id, u32 seconds)
 			      interval);
 }
 
-void upf_pfcp_server_deferred_release_association (u32 node)
+void upf_pfcp_server_deferred_free_msgs_by_node (u32 node)
 {
-  upf_main_t *gtm = &upf_main;
   pfcp_server_main_t *psm = &pfcp_server_main;
+  hash_set (psm->free_msgs_by_node, node, 1);
+}
 
-  if (!pool_is_free_index (gtm->nodes, node))
-    {
-      /*
-       * Delay releasing the association till the end of the loop
-       * to avoid problems with stopping just expired timers
-       */
-      vec_add1 (psm->release_node_assoc, node);
-    }
+void upf_pfcp_server_deferred_free_msgs_by_sidx (u32 sidx)
+{
+  pfcp_server_main_t *psm = &pfcp_server_main;
+  hash_set (psm->free_msgs_by_sidx, sidx, 1);
 }
 
 void upf_server_send_heartbeat (u32 node_idx)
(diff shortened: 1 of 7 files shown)


upf/upf_pfcp_server.c
```diff
@@ -1052,19 +1057,15 @@ u32 upf_pfcp_server_start_timer (u8 type, u32 id, u32 seconds)
 			      interval);
 }
 
-void upf_pfcp_server_deferred_release_association (u32 node)
+void upf_pfcp_server_deferred_free_msgs (u32 node, u32 sidx)
```
RoadRunnr, line 1060: does it really make sense to combine the node and sidx release into one function? There is no caller that releases both at the same time.

upf/upf_pfcp_server.h
```diff
@@ -209,11 +212,14 @@ pfcp_msg_pool_add (pfcp_server_main_t * psm, pfcp_msg_t * m)
   msg = pfcp_msg_pool_get (psm);
   clib_memcpy_fast (msg, m, sizeof (*m));
   msg->is_valid_pool_item = 1;
+  msg->loc[0] = 0;
   return msg;
 }
 
+#define pfcp_msg_pool_put(psm, m) do { snprintf((m)->loc, 128, "%s:%d %s", __FILE__, __LINE__, __FUNCTION__);  _pfcp_msg_pool_put(psm, m); } while (0)
```
RoadRunnr, line 219: shouldn't that be restricted to debug builds only?

upf/upf_pfcp_server.c
ivan4th: Okay, will make it 2 functions

upf/upf_pfcp_server.h
ivan4th: That was not meant to be committed at all, thanks for spotting it! On the second thought, maybe indeed we should keep that for debug builds. Bugs such as one that PR fixes are usually rare and not always easy to reproduce, and if another problem like this one is hit, we better see how exactly it happened on the first occurrence.

Will fix, thanks

RoadRunnr (approved):
LGTM

PR #56, opened 2021-01-05
Update pfcp.c

LazyHippogriff:
Changed the description of apn_dnn. The way it is currently defined(SIMPLE_IE instead of SIMPLE_IE_FREE) leads to memory leak.


RoadRunnr (approved):
LGTM

Issue #58, opened 2021-01-06
Clean up proxy sessions together with flows


ivan4th:
Right now, there's no proper "sync" between proxy sessions and the flows.
A flow can be deleted upon a timeout while there's still a proxy session entry for that flow. The unused proxy sessions may be kept indefinitely. Also, a flow can "respawn" due to packets sent by the proxy after the flow has expired.
There is a check for flow index pointing to a non-free flowtable entry e.g. in `proxy_start_connect_fn()`, but that is not correct either as that function can be invoked after quite a bit of time after the entry has been removed, in which case there'll be another flow in its place, leading to undefined behavior.

Trying to delete a proxy session right away poses some problems, namely, it's hard to avoid "late" callbacks from the VPP session layer, e.g. it's very hard to get rid of pending half-open connections properly (e.g. active connections in `TCP_STATE_SYN_SENT` state).

Suggested solution: add 'flow expiration' hook. When a flow is about to expire, the hook is invoked, and if it returns `0`, the flow is removed. Otherwise, the flow expiration is rescheduled.

The flows should contain proxy session index instead of connection index that it has now. If there's no proxy session associated with the flow, the proxy session index contains `~0`.

The proxy installs the expiration hook that checks if the flow has proxy session index other than `~0`. If so, it "nudges" the proxy session via `proxy_session_try_close_unlocked()` (with proper locking around it), and returns `-1` to tell the flowtable that the flow can't be removed just yet. `proxy_session_free()` sets proxy session index to `~0` in the corresponding flow, after which 'flow expiration hook' will succeed.

Alternatives: 'refcount' for flows. Perhaps not really necessary till only proxy sessions keep the flows from being freed. Also, this doesn't cover 'nudging' the proxy to clean up its session.
Also: not using the hook, just calling a proxy function from flowtable. But that's too much coupling IMO.

Issue #51, opened 2020-12-28
Memory leak when using proxy


ivan4th:
There's a memory leak in UPG when using proxy, possibly mostly for non-stitched TCP connections.
The leak is not visible in `show memory`, but vpp RSS keeps increasing.
It is likely related to SVM FIFOs being not properly freed.

Below is a flame graph constructed from page fault tracing with perf (based on [1]).
<img width="1197" alt="Flame Graph" src="https://user-images.githubusercontent.com/190448/103213588-352eb780-491f-11eb-9c8e-aa8ef32ec9ce.png">

[1] http://www.brendangregg.com/FlameGraphs/memoryflamegraphs.html

RoadRunnr:
I'm not convinced that the memory is indeed leaking. In order to archive decent performance, the socket buffers per connection have to be quite large. Some megabyte per connection are reasonable. The total memory required for all sockets on a medium load can easily top multiple gigabytes.

ivan4th:
The leak is no longer observed as of v1.1.4 (after #67 was merged before an earlier release)

PR #59, opened 2021-01-06
Fix failed active proxy connections not being freed

ivan4th:
Related to #51 (but also need to kill proxy sessions together with flows, see #58)


RoadRunnr (approved):
LGTM

PR #63, opened 2021-01-08
set Rx IfIndex to the NWI interface in FAR processing

RoadRunnr:
Set the `[VLIB_RX]` buffer metadata (the ingress sw_if_index) during
forwarding to the interface of the outgoing Network Instance.
Together with the setting of the FIB table index, that ensures that
further processing of the packet (e.g. NAT) is happening in the
current context.


ivan4th (approved):

ivan4th:
LGTM

Issue #61, opened 2021-01-06
Flows don't expire when there's low traffic


ivan4th:
The flows may take a very long time to expire if packets don't go through UPG every second.

PR #67, opened 2021-01-14
Fix flow expiration

ivan4th:
* Fix an issue with flow expiration during low traffic (#61)
* Fix flows not expiring due to "time going backwards" (bad current time being used in the code)
* Remove proxy sessions together with flows, after closing them (#58)
* Improve flow diagnostics for debug builds (use flow indices)
* Fix the names of flow lifetime setting functions (there was an unresolved reference in the CLI code)
* Add e2e test for flow expiration

This PR should also clean some possibly remaining proxy session leaks (#51)

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"time"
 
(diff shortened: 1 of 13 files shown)


RoadRunnr (approved):
C part looks good to me.

sergeymatov (approved):

PR #68, opened 2021-01-18
Avoid redefinition for clib_debug > 1

sergeymatov:
flowtable_main_t *fm already present in function signature.
Building with CLIB_DEBUG > 1 causes redefinition. Assuming
flowtable_main_t *fm is a signleton structure removing of
redefinition is safe.


RoadRunnr (approved):
LGTM

ivan4th (approved):

PR #69, opened 2021-01-21
Add worker support

ivan4th:
This commit makes UPG usable with at least 1 worker thread. PFCP runs on the main thread in this case, and all of the user traffic is handled by the workers.

Details of the UPG changes:

### Fix 1s delay before sending PFCP responses

There's a problem with session queue node not being run in case if there's > 0 workers. The culprit is here:
https://github.com/FDio/vpp/blob/09332e9bc09748dd7eea564cb47f70863374ebf8/src/vnet/session/session.c#L1765-L1787
As it can be seen, there's `poll-main` option introduced in newer VPP, that fixes this problem:
https://github.com/FDio/vpp/commit/e92c946498f544d6d435353eb06aa07fb20e054e
Yet in the version we use, that option is not available.
The trick we use is this:
```
  if (vm->thread_index == 0 && vlib_num_workers ())
    session_queue_run_on_main_thread (vm);
```
The same construct is also used in VPP TCP stack.

### Fix dumping the flows with multiple cores used

The flows were only being shown for the 0th core (main core), and as all of the traffic goes to the worker, that mean that `show upf flows` didn't work. This is fixed (although may need some improvements re thread safety).

### Fix cpu_index tracking (debug only) in the flowtable

This doesn't affect the functionality but could cause crashes in debug images.

### Fix proxy init

Possibly can be done in a better way. The proxy init was crashing without this reshuffling.

### E2E fixes

`make e2e-{debug|release} E2E_MULTICORE=1 ...` runs e2e tests in the multicore mode. In this mode,
E2E tests adjusted to use interface placement to put Sx on the main thread. Single worker thread is added.
Before the tests start, the runner measures CPU load for 5s, and picks the least loaded cores first of all, allocating 2 cores per each of the parallel test cases (just 2 cores in total in non-parallel mode).
For non-multicore mode, just 1 core is allocated for each of the parallel test cases.

## Known problems

**UPDATE: the race below is fixed.** There are still races in the code but they're not triggered by current E2E tests.

There appears to be a race condition in session modification handling, so far hit once in this E2E test:
`PGW IPv6 PDR replacement [proxy off-on+redirects] [It] doesn't permanently disrupt UPG's HTTP redirects`
Apparently, PDR went away at a wrong moment, although we do use worker barriers when modifying the session.
It is not currently clear whether this is multicore-only race or if we have missed some corner case in session modification handling.
```
Thread 1 "vpp_main" received signal SIGSEGV, Segmentation fault.
proxy_start_connect_fn (session_index=0x13008a851) at /src/vpp/src/plugins/upf/upf_proxy.c:374
374	/src/vpp/src/plugins/upf/upf_proxy.c: No such file or directory.
#0  proxy_start_connect_fn (session_index=0x13008a851) at /src/vpp/src/plugins/upf/upf_proxy.c:374
#1  0x00007ffff7bbb7c4 in vl_api_rpc_call_t_handler (mp=0x13008a838) at /src/vpp/src/vlibmemory/vlib_api.c:531
#2  0x00007ffff7bc4b91 in vl_msg_api_handler_with_vm_node (am=0x7ffff7dd1c80 <api_global_main>, vlib_rp=<optimized out>, the_msg=0x13008a838, vm=<optimized out>, node=0x7fffb482bd00, is_private=0 '\000') at /src/vpp/src/vlibapi/api_shared.c:635
#3  0x00007ffff7ba6d88 in vl_mem_api_handle_rpc (vm=0x7ffff5dc66c0 <vlib_global_main>, node=0x7fffb482bd00) at /src/vpp/src/vlibmemory/memory_api.c:746
#4  0x00007ffff7bb575b in vl_api_clnt_process (vm=0x7ffff5dc66c0 <vlib_global_main>, node=0x7fffb482bd00, f=<optimized out>) at /src/vpp/src/vlibmemory/vlib_api.c:337
#5  0x00007ffff5b4c1a7 in vlib_process_bootstrap (_a=<optimized out>) at /src/vpp/src/vlib/main.c:1454
#6  0x00007ffff4f92ec4 in clib_calljmp () from /builds/travelping/upg-vpp/vpp/build-root/install-vpp-native/vpp/lib/libvppinfra.so.20.09
#7  0x00007fffb4b86550 in ?? ()        
#8  0x00007ffff5b41cd2 in vlib_process_startup (vm=0x7ffff5dc66c0 <vlib_global_main>, p=0x7fffb482bd00, f=0x0) at /src/vpp/src/vlib/main.c:1479
#9  dispatch_process (vm=0x7ffff5dc66c0 <vlib_global_main>, p=0x7fffb482bd00, f=0x0, last_time_stamp=<optimized out>) at /src/vpp/src/vlib/main.c:1535
```

### Deployment notes

In case of multiple workers, we may need to do extra tricks like checking cpuset cgroup info to see which logical CPU cores are currently available and put these into the VPP startup config as `main-core` and `corelist-workers` (just 1 core in `corelist-workers`)


upf/flowtable_init.c
```diff
@@ -65,15 +65,13 @@ flowtable_max_lifetime_update (u16 value)
 }
 
 static clib_error_t *
-flowtable_init_cpu (flowtable_main_t * fm, flowtable_main_per_cpu_t * fmt)
+flowtable_init_cpu (flowtable_main_t * fm, u32 cpu_index)
```
RoadRunnr, line 68: I don't see why this change is needed.

upf/upf_cli.c
```diff
@@ -728,6 +730,8 @@ upf_gtpu_endpoint_add_del_command_fn (vlib_main_t * vm,
   int rv;
   u8 *s;
 
+  upf_proxy_init (vm);
```
RoadRunnr, line 733: why does this cli command the proxy to be ready? I don't see that is has a relation to the proxy process.
And if it does, then the call need to go into vnet_upf_upip_add_del.

upf/upf_cli.c
```diff
@@ -455,6 +455,8 @@ upf_tdf_ul_table_add_del_command_fn (vlib_main_t * vm,
   if (table_id == ~0)
     return clib_error_return (0, "table-id must be specified");
 
+  upf_proxy_init (vm);
```
RoadRunnr, line 458: why does this cli command the proxy to be ready? I don't see that is has a relation to the proxy process.
And if it does, then the call need to go into vnet_upf_tdf_ul_table_add_del.

upf/flowtable_init.c
ivan4th: In debug builds, this function sets `f->cpu_index = cpu_index;` for the flows in the flow entry cache. Before this change, it was erroneously using `os_get_thread_index()` to get that `cpu_index`, which is incorrect b/c this init procedure runs on the main thread, but initializes the per-cpu flowtable structures for all of the cores. Thus, we were getting assertions error later with flows having invalid `cpu_index`. 

Of course, we can keep `flowtable_main_per_cpu_t * fmt` argument and then get proper CPU index like this`u32 cpu_index = fmt - fm->per_cpu`, but I thought that first getting the pointer from the index in `flowtable_init()` and then getting index again from the pointer is a bit excessive.

upf/upf_cli.c
ivan4th: Fixed proxy init, now this is no longer needed and thus removed

upf/upf_cli.c
ivan4th: Fixed proxy init, now this is no longer needed and thus removed

RoadRunnr (approved):
C part LGTM

sergeymatov (approved):
LGTM

PR #73, opened 2021-02-02
[WiP] Fix leak in ip rules code and don't crash on empty apps

ivan4th:


upf/upf_app_db.c
```diff
@@ -223,6 +223,9 @@ upf_adf_get_adr_db (u32 application_id, acl_rule_t ** acl)
   upf_main_t *sm = &upf_main;
   upf_adf_app_t *app;
 
+  if (acl)
+    *acl = 0;
+
```
RoadRunnr, line 228: IMHO, the caller should take care of that

PR #74, opened 2021-02-04
Add shared IP application rules

ivan4th:
This PR makes IP application rules shared across session by using separate per-app FIB tables with fake DPOs. Inspired by [adl](https://github.com/FDio/vpp/tree/master/src/plugins/adl) and [gbp](https://github.com/FDio/vpp/tree/master/src/plugins/gbp) VPP plugins.

The optimization is only done for `from ...` part of the IP filter expression which is used to specify the target subnets/IPs in the application rule definition commands.

This PR also adds a unit test for IP rule matching (separate FIB tables).

- [X] make IP rules shared
- [X] unit test
- [X] e2e tests for IP rules
- [X] make sure IP rules work for non-proxy apps
- [x] add downstream VPP patch that fixes `ping` (needed for IP rule e2e tests)
- [x] check if `build_pfcp_rules()` can be simplified (some parts of it are probably redundant)

Possibly for a followup:

- [ ] verify that IP rules specified for proxy apps work for non-TCP (non-proxied) traffic
- [ ] verify that IP rules specified for proxy apps work for TCP traffic
- [ ] verify IP rules with intersecting IPs/subnets but different port ranges
- [ ] free empty unused FIB tables
- [ ] don't rebuild app db upon each rule creation

### DIFF
--- a/upf/unittest.c
+++ b/upf/unittest.c
@@ -1,0 +1,299 @@
+#include <vlib/vlib.h>
+
+#include "upf.h"
+#include "upf_ipfilter.h"
+#include "upf_app_db.h"
+
+static int upf_test_do_debug = 0;
+
+/* based on fib_test.c */
+#define UPF_TEST_I(_cond, _comment, _args...)		\
+  ({							\
+    int _evald = (_cond);				\
+    if (!(_evald)) {					\
+      fformat(stderr, "FAIL:%d: " _comment "\n",	\
+	      __LINE__, ##_args);			\
+      res = 1;						\
+    } else {						\
+      if (upf_test_do_debug)				\
+	fformat(stderr, "PASS:%d: " _comment "\n",	\
+		__LINE__, ##_args);			\
+    }							\
+    res;						\
+  })
+#define UPF_TEST(_cond, _comment, _args...)	\
+  {						\
+    if (UPF_TEST_I(_cond, _comment, ##_args))	\
+      ASSERT(!("FAIL: " _comment));		\
+  }
+
+static int
+test_add_ip_rule (u8 *app_name, u32 id, char *ipfilter)
+{
+  int res = 0;
+  upf_main_t * gtm = &upf_main;
+  unformat_input_t input;
+  acl_rule_t rule;
+  unformat_init_string (&input, ipfilter, strlen(ipfilter));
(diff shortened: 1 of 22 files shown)


upf/unittest.c
```diff
@@ -0,0 +1,299 @@
...
+
+  /* FIXME: trying to remove FIB table causes crash
+  UPF_TEST (fib_table_find (FIB_PROTOCOL_IP4, app_table_id) == ~0,
+            "IPAPP FIB table removed (IP4)");
+  UPF_TEST (fib_table_find (FIB_PROTOCOL_IP6, app_table_id) == ~0,
+            "IPAPP FIB table removed (IP6)");
+  */
+
+  UPF_TEST (upf_app_add_del (gtm, app_name_any, 0, 0) == 0, "del IPANY");
+  /* FIXME: trying to remove FIB table causes crash
```
sergeymatov, line 118: crash caused by assertion of FIB being referenced?
If so, I suppose to skip this case and remove commented code, while leaving just note
If not, and test case assumes FIB to be clear and ready for deletion, probably we'll have to dig into

upf/unittest.c
```diff
@@ -0,0 +1,299 @@
...
+}
+
+/* *INDENT-OFF* */
+VLIB_CLI_COMMAND (test_upf_command, static) =
+  {
+    .path = "test upf",
+    .short_help = "upf unit test",
+    .function = test_upf_command_fn,
+  };
+/* *INDENT-ON* */
```
sergeymatov, line 294: Should we move this to upf_cli.c file?

upf/upf_app_db.h
```diff
@@ -54,12 +55,35 @@ CLIB_PACKED (struct tls_client_hello_hdr
 	     u8 random[32];	/* gmt_unix_time + random_bytes[28] */
 	     });
 
+typedef struct
+{
+  u32 rule_index;		/* referenced ADR rule index */
+  index_t next;			/* link to the next less specific ACL ref */
```
sergeymatov, line 61: indent of clarification note :)

sergeymatov (approved):
I have several minor concerns - most of them about commented code and CLI moving to upf_cli file
in overal lgtm

PR #76, opened 2021-02-10
Fix non-recoverable AF_PACKET TX error

ivan4th:
Add a part of upstream AF_PACKET fix.
The full patch breaks E2E tests, so only a part of it is used that
fixes the stuck TX error and improves diagnostics.

Ref: https://gerrit.fd.io/r/c/vpp/+/31078


sergeymatov (approved):
LGTM

mgumz:
the subject should be something like "Fix AF_PACKET behaviour xyz" … coz then the commit subject is stating what the patch provides feature/fix wise.

PR #79, opened 2021-02-17
Update VPP to 21.01

ivan4th:
Update vpp.spec
Update the build image tag
Update vpp-patches removing those already present in the upstream
Update pool_foreach usage
Fix SVM FIFO setup in the proxy
Fix indent issues

Initial migration done by @RoadRunnr


RoadRunnr (approved):
LGTM

one remark though, the build is running on Ubuntu focal. That means the tricks to get a correct indent package version are no longer required. Might be sensible to remove that in another PR.

sergeymatov (approved):

PR #81, opened 2021-02-17
Add downstream 'set ip neighbor-config' patch

ivan4th:
This is a backport of the patch for VPP 21.01, to be posted / merged after #79 is merged.

It adds `set ip neighbor-config` command:

```
set ip neighbor-config ip4|ip6 [limit <limit>] [age <age>] [recycle|norecycle]
```

Setting `age` (in seconds) makes VPP re-issue ARP requests for the expired entries once that 'age' passes:
```
17:29:43.336204 ARP, Request who-has 10.0.1.3 tell 10.0.1.2, length 28
17:29:43.336225 ARP, Reply 10.0.1.3 is-at 8a:ee:05:35:c7:99, length 28
```


ivan4th:
Err wrong target branch

PR #85, opened 2021-02-22
vpp-patches: add 'set ip neighbor-config' patch

ivan4th:
Upstream: https://gerrit.fd.io/r/c/vpp/+/31363


sergeymatov (approved):
Looks good to me

PR #88, opened 2021-02-25
Detect duplicate SEIDs in Session Establishment Requests

ivan4th:
The Session Establishment Requests with duplicate SEIDs are now be rejected with "Request Rejected" (64) PFCP Cause Value and a TP-specific Error Report IE.

Includes commit from #86 which needs to be merged before this PR.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -193,7 +193,9 @@ func (f *Framework) AfterEach() {
 		if f.PFCP != nil {
 			// FIXME: we need to make sure all PFCP packets are recorded
 			time.Sleep(time.Second)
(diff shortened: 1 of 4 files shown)


mgumz:
@ivan4th looks ok in general. two remarks: please put a comment in there which states that the SEID is actually per SMF and that an overhaul of this part of the code is planned in near term future. 2nd: do you know the performance impact of pcfp_lookup(seid)?

ivan4th:
Will add the comment. Concerning `pfcp_lookup()`: this is a quick hash lookup, so I don't think it'll have noticeable negative performance impact, especially given that session creation can take a few milliseconds anyway

sergeymatov (approved):

PR #91, opened 2021-02-26
vpp-patches: Avoid simple counter overflow

sergeymatov:
upstream patch https://gerrit.fd.io/r/c/vpp/+/31433


mgumz:
@sergeymatov is this fixing something critical? i can't deduce the impact by reading the commit message.

sergeymatov:
> @sergeymatov is this fixing something critical? i can't deduce the impact by reading the commit message.

it's not
we've faced issue with counter overflow once, and tbh - it's just an improvement for counters.
BUT measuring amount of heat around association counter being equal u64_max for a single time, I would suggest that we've never face such thing anymore

mgumz:
what kind of improvement is it?

PR #99, opened 2021-03-19
e2e: improve backtrace output during crashes

ivan4th:
Use `bt full 30` instead of `bt 10` in gdb to check where VPP did crash


jellonek (approved):
LGTM

RoadRunnr (approved):

Issue #87, opened 2021-02-22
"Out of memory" crash in PFCP Association Release E2E test


ivan4th:
So far the problem was only observed in the CI environment. It is not clear at this point if it's a test issue or a real bug. Lowering the rate of session modifications appears to make this problem more rare.
This problem appears only to happen with release builds.

Test:
```
[Fail] Clearing message queue during PFCP Association Release upon timeout [It] should work correcty
```

Diagnostics:
```
Program received signal SIGABRT, Aborted.
0x00007ffff5cb618b in raise () from /lib/x86_64-linux-gnu/libc.so.6
#0  0x00007ffff5cb618b in raise () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007ffff5c95859 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007ffff6385d86 in os_panic () at /src/vpp/src/vppinfra/unix-misc.c:177
#3  0x00007ffff6389637 in clib_mem_alloc_aligned_at_offset (size=2056, align=<optimized out>, align_offset=<optimized out>, os_out_of_memory_on_failure=1) at /src/vpp/src/vppinfra/mem.h:243
#4  vec_resize_allocate_memory (v=<optimized out>, length_increment=2048, data_bytes=2056, header_bytes=8, data_align=<optimized out>, numa_id=<optimized out>) at /src/vpp/src/vppinfra/vec.c:67
#5  0x00007fffb3258a45 in _vec_resize_inline (v=0x0, length_increment=<optimized out>, data_bytes=<optimized out>, header_bytes=<optimized out>, data_align=<optimized out>, numa_id=<optimized out>) at /src/vpp/src/vppinfra/vec.h:170
#6  upf_pfcp_make_response (resp=0x7ffff3f04580, req=0x7fffefaf9ec0, len=2048) at /src/vpp/src/plugins/upf/upf_pfcp_server.c:520
#7  upf_pfcp_send_response (req=0x7fffefaf9ec0, cp_seid=18299514082375153098, type=<optimized out>, grp=<optimized out>) at /src/vpp/src/plugins/upf/upf_pfcp_server.c:532
#8  0x00007fffb32531e3 in session_msg (msg=<optimized out>) at /src/vpp/src/plugins/upf/upf_pfcp_api.c:2644
#9  upf_pfcp_handle_msg (msg=<optimized out>) at /src/vpp/src/plugins/upf/upf_pfcp_api.c:181
```

[...]

PR #100, opened 2021-03-19
Add downstream fix for UDP buffer chaining

ivan4th:
Upstream CR: https://gerrit.fd.io/r/c/vpp/+/31647 (merged)
vpp-dev thread: https://lists.fd.io/g/vpp-dev/message/18921

May provide a stable backport if we decide it's necessary

### DIFF
--- a/vpp-patches/0013-session-fix-populating-chained-buffers.patch
+++ b/vpp-patches/0013-session-fix-populating-chained-buffers.patch
@@ -1,0 +1,42 @@
+From bbee13b6f04515ee42f066f3f0bfc6b675b5a9a3 Mon Sep 17 00:00:00 2001
+From: Ivan Shvedunov <ivan4th@gmail.com>
+Date: Mon, 15 Mar 2021 19:05:14 +0300
+Subject: [PATCH] session: fix populating chained buffers
+
+Type: fix
+
+Signed-off-by: Ivan Shvedunov <ivan4th@gmail.com>
+Change-Id: I3fdc1711d508bc5b0c2f56b3bd3fb4a56056eb54
+---
+ src/vnet/session/session_node.c | 8 +++++---
+ 1 file changed, 5 insertions(+), 3 deletions(-)
+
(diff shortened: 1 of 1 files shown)


vpp-patches/0013-session-fix-populating-chained-buffers.patch
```diff
@@ -0,0 +1,42 @@
...
+ 	      session_dgram_hdr_t *hdr = &ctx->hdr;
+ 	      u16 deq_now;
++	      u32 offset;
++
+ 	      deq_now = clib_min (hdr->data_length - hdr->data_offset,
+ 				  len_to_deq);
+-	      n_bytes_read = svm_fifo_peek (f, hdr->data_offset, deq_now,
+-					    data);
++	      offset = hdr->data_offset + SESSION_CONN_HDR_LEN;
++	      n_bytes_read = svm_fifo_peek (f, offset, deq_now, data);
```
RoadRunnr, line 29: Upstream really needs to judge this. I guess it is good enough for us for the moment.

Anyhow, wouldn't this be enough and a much smaller change?
```
	      n_bytes_read = svm_fifo_peek (f, hdr->data_offset + SESSION_CONN_HDR_LEN,
					    deq_now, data);
```

RoadRunnr (approved):

vpp-patches/0013-session-fix-populating-chained-buffers.patch
ivan4th: I copied the construct from the other part of the code which does similar thing and made it look the same b/c if you have to do some repetition in the code (and in case of VPP that's rather often the case), it better should look as similar as possible. I could have changed the other place too instead, but maybe they did it just to make the lines shorter, so indent doesn't mangle them too much, and compiler should eliminate the difference

ivan4th:
The e2e failure is caused by #87 which is fixed by #102, I suggest that we just merge this PR instead of re-running e2e another time

PR #107, opened 2021-03-31
Switch to GitHub actions

ivan4th:
This PR switches the repository to using GitHub Actions instead of GitLab CI.
The workflow uses a self-hosted runner on a k8s cluster to execute the jobs. In order to avoid possible troubles with DIND, the buildenv is executed directly on the cluster via kubectl. The new k8s buildenv mode can also be used to utilize an existing k8s cluster to build and test upg-vpp during development.

### TODO
- [x] switch to GitHub Actions
- [x] slack notification
- [x] [changelog](https://github.com/mikepenz/release-changelog-builder-action), [releases](https://github.com/softprops/action-gh-release) and promoting the images
- [x] run all of E2E tests (currently E2E_FOCUS is added to make workflow debugging quicker)

### TODO in followups
- [display jUnit output](https://github.com/marketplace/actions/junit-report-action) from e2e in the annotations
- get rid of update-build-image-tag (use `FROM arg`)
- fail checkstyle if `ZZZZZ:` is found
- dockerfiles and image builds for rsync and the runner
- support stable branch
- extended nightly/master/stable only tests (ASAN, non-quick e2e, etc)

### Caveats about switching to GitHub Actions from GitLab CI
Copied from [my twitter thread](https://twitter.com/ivan4th/status/1376908128948740100).

1. You can't rerun just some jobs for your workflow, only all of the
   jobs: https://github.community/t/re-run-jobs/16145/10 Too bad if
   some of your test jobs are flaky sometimes.
2. There's no inheritance for jobs. There's build matrix that helps
   sometimes, but to a limited extent. There's also "composite run
   steps action" but it can't include other actions in it
3. Not many customizations available for docker actions, e.g. you
   can't increase shm (e.g. for a database) or run docker actions in
   privileged containers even if you use a self-hosted runner.
   https://github.com/actions/container-action/issues/2
4. In the aforementioned matrix builds, you can't customize 'needs'.
   This means that you can't make a matrix of test jobs where "test
   debug" job only depends on the debug build and "test release" job that
   only depends on the release build:
   https://github.community/t/matrix-variable-does-not-work-in-needs-keyword/16972
5. You might consider using strategy: fail-fast: false for the matrix
   jobs. The default of abandoning the matrix once one of the jobs fails
   doesn't make too much sense IMO. Especially if some of the tests are
   flaky ;(
6. A more advanced topic: if you're using docker buildx don't get too
   excited about caching using actions/cache@v2. It will cache the layers
   all right but if you're using --mount=type=cache in your Dockerfiles
   for things like ccache or apt that will not be stored
   (that's a buildkit issue though, not one in GH Actions)
7. There's no easy way to limit the parallelism of the jobs, which may
   make quite a bit of sense in case if you're using self-hosted runners.
   Not so easy way: https://github.com/marketplace/actions/action-turnstyle


RoadRunnr:
there is a `Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?` error in the prepare stage. Is that expected?

ivan4th:
The docker error in prepare is harmless (due to extra check for local docker image that's not working with external buildkit) but I will fix it, thanks

RoadRunnr (approved):

sergeymatov (approved):

PR #109, opened 2021-04-07
Don't do flow expiration checks too often

ivan4th:
The checks can be rather costly if there are many flows.
Doing them twice per second at most instead of doing the check upon
every frame handled by the flow node should be enough.

### DIFF
--- a/upf/upf.c
+++ b/upf/upf.c
@@ -322,6 +322,7 @@ upf_init (vlib_main_t * vm)
   sm->vlib_main = vm;
   sm->pfcp_spec_version = 15;
   sm->rand_base = random_default_seed ();
+  sm->next_flow_expiration = 0;
 
   if ((error = vlib_call_init_function (vm, upf_proxy_main_init)))
     return error;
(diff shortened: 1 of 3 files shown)


RoadRunnr (commented):
I'm not convinced that this is the right approach.

From a quick inspection of `flowtable_timer_expire` I get the impression that `time_slot_curr_index` is advanced for every invocation of `flowtable_timer_expire`.
That would run over the wheel far too fast and indeed cause to many entries to be checked.

It is somewhat questionable that the per CPU timer implementation is a good idea in any case. There is no guarantee that a flow is always handled by the same CPU. This could lead to race condition where one CPU is expiring the entry while another CPU tries to access then entry (somewhat unlikely, because the flow must have been idle for some time for this).

The [flowtable plugin](https://github.com/GabrielGanne/vpp-flowtable-plugin) has moved to a different timer logic. It might make sense to resync our copy with theirs.

ivan4th:
@RoadRunnr indeed it looks there's a bug in #67, that was causing this "overadvancing", will check. Without bumping `time_slot_curr_index` / `next_check` on every call, the code would still possibly check hundreds of flows upon every arriving packet, but that can probably be handled in a cleaner way, will fix.

Re newer flowtable plugin: we will switch to it as part of the multicore/multiworker effort once I get back to it.

ivan4th:
Closing in favor of #111

PR #117, opened 2021-05-05
[testing] Diagnose and fix message queue expiration issues

ivan4th:
In pfcp_process(), check whether there are any messages that should
have expired for now and delete them, logging warnings

### DIFF
--- a/upf/upf_pfcp_server.c
+++ b/upf/upf_pfcp_server.c
@@ -347,6 +347,7 @@ enqueue_request (pfcp_msg_t * msg, u32 n1, u32 t1)
   hash_set (psm->request_q, msg->seq_no, id);
   msg->timer =
     upf_pfcp_server_start_timer (PFCP_SERVER_T1, msg->seq_no, msg->t1);
+  msg->expires_at = psm->now + msg->t1;
 }
 
 static void
@@ -371,6 +372,7 @@ request_t1_expired (u32 seq_no)
       upf_debug ("resend...\n");
       msg->timer =
 	upf_pfcp_server_start_timer (PFCP_SERVER_T1, msg->seq_no, msg->t1);
+      msg->expires_at = psm->now + msg->t1;
 
       upf_pfcp_send_data (msg);
     }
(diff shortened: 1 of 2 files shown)


sergeymatov (commented):
In general it's more like a workaround for unclear behavior.
I guess we should have it only after testing is done

upf/upf_pfcp_server.c
```diff
@@ -1228,11 +1232,29 @@ static uword
       /* Free messages that must be freed after the loop */
       pool_foreach (msg, psm->msg_pool)
       {
-	if (!msg->is_valid_pool_item ||
-	    (!hash_get (psm->free_msgs_by_node, msg->node) &&
-	     !hash_get (psm->free_msgs_by_sidx, msg->session_index)))
+	if (!msg->is_valid_pool_item)
```
RoadRunnr, line 1235: AFAIK this should never happen, a assert for the debug build and clib_warning for the release images might be better.

upf/upf_pfcp_server.c
```diff
@@ -1228,11 +1232,29 @@ static uword
       /* Free messages that must be freed after the loop */
       pool_foreach (msg, psm->msg_pool)
```
RoadRunnr, line 1233: this whole loop is IMHO bogus. Instead of checking all messages, putting the indexes of the to be freed messages in a vector and the running only over that vector should be more efficient.

RoadRunnr (approved):

upf/upf_pfcp_server.c
ivan4th: Thanks, will fix that

upf/upf_pfcp_server.c
ivan4th: That's true. But the point of this PR is also to verify whether we have timer wheel issues, and checking all of the timestamps is the easiest way there. In a followup, it makes sense to use FIFOs and check timestamps starting from the head instead of using timer wheels for tracking message expiration, so this loop will be removed.

upf/upf_pfcp_server.c
ivan4th: Looking further at this problem, this is not really a bug b/c `psm->msg_pool` is used in such way that there may be invalid messages in it. `pfcp_msg_pool_put()` doesn't use `pool_put()`, but rather sets `is_valid_pool_item` to `0` and then adds the message to a separate freelist, which is later used to populate `msg_pool_cache` for faster allocation. So, going through all of the msg pool with `pool_foreach()` is indeed quite inefficient (as noted below, I will fix that in a followup) but also it can stumble upon "invalid" messages in the pool.

PR #126, opened 2021-05-24
vpp-patches: cherry-pick mem leak fix in show memory main-heap

ivan4th:
Upstream CR: https://gerrit.fd.io/r/c/vpp/+/31468


sergeymatov (approved):
LGTM

PR #129, opened 2021-06-02
Drop sessions with duplicate UE IP

ivan4th:
Drop the conflicting session upon a UE IP conflict, sending
Session Report Request for the dropped session as per
TS 29.244 clause 5.18.2: UP Function Initiated PFCP Session Release

### DIFF
--- a/upf/upf_pfcp_api.c
+++ b/upf/upf_pfcp_api.c
@@ -657,28 +657,63 @@ handle_f_teid (upf_session_t * sx, upf_main_t * gtm, pfcp_pdi_t * pdi,
 
 }
 
+/*
+ * Delete session that conflicts on UE IP if it resides
+ * on the same CP node as the new one.
+ */
+static void
+purge_conflicting_session (upf_session_t * sx_old,
+			   upf_session_t * sx_new,
+			   pfcp_ue_ip_address_t * ue_addr)
+{
+  if (sx_old->assoc.node != sx_new->assoc.node)
+    return;
+
+  clib_warning ("UE IP conflict: deleting session CP SEID=0x%016" PRIx64
+		" %U " "due to attempt to create CP SEID=0x%0x16" PRIx64,
+		sx_old->cp_seid, format_ue_ip_address, ue_addr,
+		sx_new->cp_seid);
+
+  upf_pfcp_session_up_deletion_report (sx_old);
+
+  if (pfcp_disable_session (sx_old) != 0)
+    clib_warning ("PFCP Session %" PRIu64 " could no be disabled\n",
+		  sx_old->cp_seid);
+  else
+    pfcp_free_session (sx_old);
+}
+
 static int
-validate_ue_ip (upf_session_t * sx, upf_nwi_t * nwi,
-		pfcp_ue_ip_address_t * ue_addr)
+resolve_ue_ip_conflicts (upf_session_t * sx, upf_nwi_t * nwi,
+			 pfcp_ue_ip_address_t * ue_addr)
 {
   upf_main_t *gtm = &upf_main;
   const dpo_id_t *dpo;
+  int r = 0;
 
   if (ue_addr->flags & IE_UE_IP_ADDRESS_V4)
     {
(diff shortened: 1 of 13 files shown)


upf/upf_pfcp_api.c
```diff
@@ -657,28 +657,60 @@ handle_f_teid (upf_session_t * sx, upf_main_t * gtm, pfcp_pdi_t * pdi,
...
   upf_main_t *gtm = &upf_main;
   const dpo_id_t *dpo;
+  int r = 0;
 
   if (ue_addr->flags & IE_UE_IP_ADDRESS_V4)
     {
       dpo = upf_get_session_dpo_ip4 (nwi, &ue_addr->ip4);
       if (dpo && dpo->dpoi_index != sx - gtm->sessions)
-	return -1;
+	{
```
RoadRunnr, line 695: a real problem only exists if the same CP node tries to use the same UE IP twice. If the UE IP was allocated by a different node, that other node would still be aware of the assignment and there would be no further problems.

The function should therefore check for the node_id and only remove the other session if the node_ids match.

upf/upf_pfcp_api.c
```diff
@@ -2505,7 +2551,7 @@ handle_session_modification_request (pfcp_msg_t * msg,
   if (r == 0)
     resp->cause = PFCP_CAUSE_REQUEST_ACCEPTED;
 
-  upf_pfcp_send_response (msg, &resp_dmsg);
+  upf_pfcp_send_response (msg, &resp_dmsg, outlives_session);
```
RoadRunnr, line 2554: there is on thing that bugs me about the `outlives_session` argument, why do we keep a reference between session and messages in the response cache and request queue anyway?
There should be no reason to remove a message from the response cache even when the session is deleted. A possible resent of the response does not need to access the session in any way.
The request queue is a bit more complex,
any answer should remove the message from the queue, but it should not attempt to access the session if the session is already gone.

RoadRunnr (changes requested):

upf/upf_pfcp_server.c
```diff
@@ -517,6 +517,53 @@ urr_check_counter (u64 bytes, u64 consumed, u64 threshold, u64 quota)
...
+
+  SET_BIT (req->grp.fields, SESSION_REPORT_REQUEST_PFCPSRREQ_FLAGS);
+  /* PSDBU = PFCP Session Deleted By the UP function */
+  req->pfcpsrreq_flags = PFCPSRREQ_PSDBU;
+
+  /*
+   * outlives_session=true so that this request isn't dropped
+   * from the response queue after the session is deleted,
+   * and can be retransmitted if it's lost
+   */
```
RoadRunnr, line 562: that comment is obsolete now.

upf/upf_pfcp_api.c
```diff
@@ -657,28 +657,60 @@ handle_f_teid (upf_session_t * sx, upf_main_t * gtm, pfcp_pdi_t * pdi,
 
 }
 
+static void
+expunge_conflicting_session (upf_session_t * sx_old,
```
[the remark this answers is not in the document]
ivan4th: Fixed

upf/upf_pfcp_api.c
ivan4th: Fixed. If node id is different, the new Session Establishment Request will still fail, but the old session will not be dropped.

upf/upf_pfcp_api.c
ivan4th: Removed the code that was dropping session-related messages upon session deletion, and updated the tests accordingly.

upf/upf_pfcp_server.c
ivan4th: Removed.

ivan4th:
Had to include some not-completely-related e2e changes that I accumulated (PFCP request pipelining) so as to avoid serious merge conflicts in the E2E tests later.

sergeymatov (approved):

RoadRunnr (approved):

PR #135, opened 2021-06-09
[UPG NAT] Don't reject duplicate FARs with NAT for same UE IP

sergeymatov:


upf/upf_pfcp_api.c
```diff
@@ -1403,6 +1403,11 @@ handle_nat_binding_creation (upf_session_t * sx, u8 * nat_pool_name,
   upf_nat_addr_t *ap;
   int rc = 0;
 
+  // We already created NAT Binding for a session
+  // Using different FAR
```
ivan4th, line 1407: Please use `/* ... */` style comments ... Somehow, our style checker doesn't detect it but perhaps it would be better to have uniform style anyway.

ivan4th (approved):
LGTM

RoadRunnr (approved):

PR #139, opened 2021-06-14
[Stable] [Backport] [UPG-NAT] Per pool stats binary API

sergeymatov:
Binapi used to claim following statistics per NAT pool:
  - NAT Pool name
  - Network instance name
  - Port Block size
  - Maximum number of bindings
  - Current amount of allocated bindings


ivan4th (approved):
LGTM

PR #141, opened 2021-06-15
E2E and build improvements

ivan4th:
* add E2E_KEEP_ALL_ARTIFACTS to keep artifacts for successful e2e
* add optional UPG_BUILDENV_NODE env var for k8s mode buildenv
* e2e: don't use -stream option for Ginkgo
* fix env var name for defaulting in hack/ci-build.sh


sergeymatov (approved):
LGTM

PR #143, opened 2021-06-16
[Stable] [Backport] Fix memleak in BBF FAR processing and port range calculation for binding

sergeymatov:
- Free duplicated BBF NAT Port Block entity during FAR handling
- Correct end of port range for binding calculation


ivan4th (approved):
LGTM

PR #152, opened 2021-06-29
fix bit_rate (MBR, GBR) en/decoding

### LINKED ISSUE #150: [Help] MBR values
I was trying to interconnect the upg-vpp to free5gc which worked fine after some adaptations (and your support). However I wanted to manage the create QER request within the upg but I have an inconsistency with the MBR. When I display it on the CLI (show upf session), it is different from the MBR on the PDU session request. I would like to know how this IE is managed in the upg (encoding etc...). Thank you in advance for any clarification you can give me.

PDU session trace and show upf session command output are attached
[Trace.zip](https://github.com/travelping/upg-vpp/files/6709304/Trace.zip)

RoadRunnr:
The values are actually 40bit (5byte) wide.

resolves #150

### DIFF
--- a/upf/pfcp.c
+++ b/upf/pfcp.c
@@ -289,6 +289,27 @@ format_pfcp_msg_hdr (u8 * s, va_list * args)
     (V) += sizeof(u32);				\
     ntohl(*_V); })
 
+#define put_u40(V,I)					\
+  do {							\
+    (V)[_vec_len((V))] = (I) >> 32;			\
+    (V)[_vec_len((V)) + 1] = ((I) >> 24) & 0xff;	\
+    (V)[_vec_len((V)) + 2] = ((I) >> 16) & 0xff;	\
+    (V)[_vec_len((V)) + 3] = ((I) >> 8) & 0xff;		\
+    (V)[_vec_len((V)) + 4] = (I) & 0xff;		\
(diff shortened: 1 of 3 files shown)


upf/pfcp.c
```diff
@@ -1095,11 +1110,11 @@ decode_bit_rate (u8 * data, u16 length, void *p)
 {
   pfcp_bit_rate_t *v = p;
 
-  if (length < 8)
+  if (length < 10)
     return PFCP_CAUSE_INVALID_LENGTH;
 
-  v->ul = get_u32 (data);
-  v->dl = get_u32 (data);
+  v->ul = get_u40 (data);
```
ivan4th, line 1116: This is causing build error
```
src/vpp/src/plugins/upf/pfcp.c:1116:11: error: shift count >= width of type [-Werror,-Wshift-count-overflow]
```
apparently due to `data` being `u8 *` upon `(V)[0] << 32` etc

ivan4th: Ref: https://github.com/travelping/upg-vpp/runs/2944778616?check_suite_focus=true#step:3:3127

RoadRunnr: fixed, but I really wonder why this does not happen for u24 getter.

ivan4th (approved):
LGTM

PR #153, opened 2021-07-01
[UPG NAT] Fix output FIB index selection for controlled NAT

sergeymatov:
For multi-NWI setup, NAT function vanilla mechanism of output FIB selection is not working correctly if controlled NAT function is used. Outisde FIB index can be used based on NWI in which NAT pool has been created.


ivan4th (approved):
LGTM

ivan4th (approved):
LGTM

PR #155, opened 2021-07-01
 [Stable] [Backport] [UPG NAT] Fix output FIB index selection for controlled NAT 

sergeymatov:
For multi-NWI setup, NAT function vanilla mechanism of output FIB selection is not working correctly if controlled NAT function is used. Outisde FIB index can be used based on NWI in which NAT pool has been created.

### DIFF
--- a/vpp-patches/0015-Controlled-NAT-function.patch
+++ b/vpp-patches/0015-Controlled-NAT-function.patch
@@ -1,4 +1,4 @@
-From b1b398a58c3823e32b422110a3d39f9447f2e82b Mon Sep 17 00:00:00 2001
+From a9d428c487150d2d5d7e5230aee38adac9ad7b9a Mon Sep 17 00:00:00 2001
 From: Sergey Matov <sergey.matov@travelping.com>
 Date: Tue, 18 May 2021 17:07:31 +0400
 Subject: [PATCH] Controlled NAT function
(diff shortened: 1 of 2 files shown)


vpp-patches/0015-Controlled-NAT-function.patch
```diff
@@ -7,32 +7,32 @@ Allow to override ED NAT address and port selection in slow path
...
-+				    u8 proto, u16 port_per_thread,
++nat_controlled_alloc_addr_and_port (snat_main_t * sm,
++				    u32 thread_index, ip4_address_t r_addr,
++				    u16 r_port, u8 proto,
 +				    u32 snat_thread_index, snat_session_t * s,
 +				    ip4_address_t * outside_addr,
 +				    u16 * outside_port,
-+				    clib_bihash_kv_16_8_t * out2in_ed_kv)
++				    clib_bihash_kv_16_8_t * out2in_ed_kv,
++				    u32 buf_out_fib_idx)
```
RoadRunnr, line 35: The argument ordering is bad. The typical rule in C is to have input arguments before output arguments in functions.
`buf_out_fib_idx` is clearly an input value.

korroot:
Closes: https://github.com/travelping/upg-vpp/issues/156

PR #158, opened 2021-07-08
Put association back in pool when released

sergeymatov:
If there are heartbeat requests for already released association
been sent from UPG yet another association release will be caused by timer.
So, assoc counter will be decreased twice.
This fix prevents secondary association release.
"show upf association" command won't show "invalid" associations anymore.

Also, fix basic e2e test to compare for an empty string


ivan4th:
LGTM

ivan4th (approved):

PR #159, opened 2021-07-08
[Stable] [Backport] Put association back in pool when released

sergeymatov:
If there are heartbeat requests for already released association
been sent from UPG yet another association release will be caused by timer.
So, assoc counter will be decreased twice.
This fix prevents secondary association release.
"show upf association" command won't show "invalid" associations anymore.

Also, fix basic e2e test to compare for an empty string

### DIFF
--- a/test/e2e/upg_e2e.go
+++ b/test/e2e/upg_e2e.go
@@ -505,7 +505,7 @@ var _ = ginkgo.Describe("Clearing message queue", func() {
 				time.Sleep(10 * time.Millisecond)
 			}
 
-			gomega.Eventually(listNodes, 2*time.Minute, 5*time.Second).Should(gomega.Equal("invalid,invalid,invalid"))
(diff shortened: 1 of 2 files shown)


mgumz:
@sergeymatov what does this PR fix? what are the symptoms of the thing it fixes?

korroot:
Closes: https://github.com/travelping/upg/issues/19
The issue is that associations are counter wrong, for some upg the associations is 0 and for others 18446744073709552000

sergeymatov:
@mgumz 
if there are heartbeat requests for already released association been sent from UPG yet another association release will be caused by timer. So, assoc counter will be decreased twice. This fix prevents secondary association release
and "show upf association" command won't show "invalid" associations with this fix

mgumz:
@sergeymatov can you please add that explanation into the commit message?

sergeymatov:
@mgumz 
done for this and for original issue

ivan4th (approved):
LGTM

PR #160, opened 2021-07-13
Fix memory leak during redirect info handling

ivan4th:


upf/upf_pfcp.c
```diff
@@ -781,13 +783,13 @@ pfcp_make_pending_far (upf_session_t * sx)
...
-	  }
-
-	new->forward.rewrite = vec_dup (old->forward.rewrite);
+	if (!(old->apply_action & FAR_FORWARD))
+	  continue;
+	if (old->forward.rewrite)
+	  new->forward.rewrite = vec_dup (old->forward.rewrite);
+	if (old->forward.flags & FAR_F_REDIRECT_INFORMATION)
+	  cpy_redirect_information (&new->forward.redirect_information,
+				    &new->forward.redirect_information);
```
RoadRunnr, line 792: The combination of the tests could leave `new->forward.redirect_information` in an inconsistent state.
Using `continue` seemed to make sense before, but with the additional copying it would make more sense to remove it.

Something like:
```suggestion
       clib_memset(&new->forward.redirect_information, 0, sizeof(new->forward.redirect_information));
	if ((old->apply_action & FAR_FORWARD) && (old->forward.rewrite))
	  new->forward.rewrite = vec_dup (old->forward.rewrite);
	if (old->forward.flags & FAR_F_REDIRECT_INFORMATION)
	  cpy_redirect_information (&new->forward.redirect_information,
				    &new->forward.redirect_information);
```

ivan4th: Hmm I just saw another problem here (`cpy_redirect_information` new to new instead of old to new), but concerning continue: doing this memset indeed makes sense, but copying old redir info in case of `!(old->apply_action & FAR_FORWARD)` should not be done, perhaps

ivan4th: from what I see in case if `FAR_FORWARD` is not set in `apply_action`, `handle_create_far` / `handle_update_far` don't populate the `forward` field at all. On the other hand, it's zeroed out so both `rewrite` and `flags & FAR_F_REDIRECT_INFORMATION` are also zero in this case. But perhaps a little defensive programming would not harm

ivan4th: So, added `clib_memset` to clear the redir info, but kept continue in place for now (below `clib_memset`)

sergeymatov (approved):
lgtm

PR #162, opened 2021-07-13
Cleanup in CLI handling

jellonek:
Code simplification by removal of 'add's, which were not even mentioned in helps (they are default actions unless `del` is specified).
Short helps formatting unification.
Fix in handling `disable` for `upf tdf ul`.


ivan4th (approved):

RoadRunnr:
the implicit and hidden add is used by many other CLI commands. I therefore don't see the need to remove it.

After all, the CLI is a debug tool for developers only.

upf/upf_cli.c
```diff
@@ -63,8 +63,6 @@ upf_pfcp_endpoint_ip_add_del_command_fn (vlib_main_t * vm,
     {
```
captain-hoity-toity, line 63: Consider using a more descriptive commit message to capture the essence of the changes.

PR #169, opened 2021-08-17
Fix buildenv startup issue on CI

ivan4th:
Use shorter devenv StatefulSet names, and, consequently, devenv pod
names.
Add 5m timeout for the devenv StatefulSet rollout.


sergeymatov (approved):
LGTM

PR #170, opened 2021-08-18
[Stable][Backport] Fix buildenv startup issue on CI

sergeymatov:
Use shorter devenv StatefulSet names, and, consequently, devenv pod
names.
Add 5m timeout for the devenv StatefulSet rollout.


ivan4th (approved):
LGTM

PR #172, opened 2021-08-18
Fix TCP stitching in the UL direction

ivan4th:
TCP sequence number offsets were not calculated properly when the
first packet being stitched was originating from the UE.


sergeymatov (approved):
LGTM

PR #173, opened 2021-08-19
Skip unstable IPv6 UDP traffic counting tests

ivan4th:
There's likely a shortcoming in E2E Linux network namespace handling
code that causes problems when binding UDP sockets, making these IPv6
test unstable. Will need to fix that later.


sergeymatov (approved):
LGTM

PR #175, opened 2021-08-23
Add iperf3-based e2e test cases and make forced TCP stitching optional

ivan4th:
This PR adds several E2E test cases that use `iperf3` to check that TCP stitching works for both UL and DL packets.
In order to make sure the problem is not masked by the current forced stitching behavior, it adds a proxy option:
```
set upf proxy force-stitching on|off
```
The status can be checked with `show upf proxy`:
```
vpp# show upf proxy 
MSS: 1250
FIFO Size: 64k
Max FIFO Size: 128m
Hi/Lo Watermark: 80 % / 50 %
Prealloc FIFOs: 0
Private Segment Count: 0
Private Segment Size: 512m
Force stitching: on
```


RoadRunnr (approved):
LGTM for the C part

Issue #186, opened 2021-10-26
UPG Crash on multi-core setup if Proxy is enabled


sergeymatov:
Packets expecting `proxy_start_connect_fn` might cause UPG crash if session for the flow has been already deleted in main thread.
Assertion (in case of debug image) or segfault could happen here.
https://github.com/travelping/upg-vpp/blob/master/upf/upf_proxy.c#L367

PR #187, opened 2021-10-26
Fix UPG proxy race conditions for multi core setup

### LINKED ISSUE #185: UPG crash on multi-core setup in proxy mode
Under high traffic rate and session churn loop UPG crashed

`#0  0x00007fcfdd50c18b in raise () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007fcfdd4eb859 in abort () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x000055e2b7a7c905 in os_exit (code=<optimized out>) at /src/vpp/src/vpp/vnet/main.c:433
#3  0x00007fcfdddd0d18 in unix_signal_handler (signum=<optimized out>, si=<optimized out>, uc=<optimized out>) at /src/vpp/src/vlib/unix/main.c:184
#4  <signal handler called>
#5  proxy_start_connect_fn (session_index=0x1300e4fe9) at /src/vpp/src/plugins/upf/upf_proxy.c:383
#6  0x00007fcfdf7fd6b4 in vl_api_rpc_call_t_handler (mp=0x1300e4fd0) at /src/vpp/src/vlibmemory/vlib_api.c:531
#7  0x00007fcfdf806b61 in vl_msg_api_handler_with_vm_node (am=0x7fcfdf814c90 <api_global_main>, vlib_rp=<optimized out>, the_msg=0x1300e4fd0, vm=<optimized out>,
    node=0x7fce5dbb1ec0, is_private=0 '\000') at /src/vpp/src/vlibapi/api_shared.c:635
#8  0x00007fcfdf7e8c48 in vl_mem_api_handle_rpc (vm=0x7fcfddded800 <vlib_global_main>, node=0x7fce5dbb1ec0) at /src/vpp/src/vlibmemory/memory_api.c:746
#9  0x00007fcfdf7f763b in vl_api_clnt_process (vm=0x7fcfddded800 <vlib_global_main>, node=0x7fce5dbb1ec0, f=<optimized out>) at /src/vpp/src/vlibmemory/vlib_api.c:337
#10 0x00007fcfddd6fc37 in vlib_process_bootstrap (_a=<optimized out>) at /src/vpp/src/vlib/main.c:1477
#11 0x00007fcfddb9cf6c in clib_calljmp () from /lib/x86_64-linux-gnu/libvppinfra.so.21.01.0
#12 0x00007fce58b8c550 in ?? ()
#13 0x00007fcfddd65462 in vlib_process_startup (vm=0x7fcfddded800 <vlib_global_main>, p=0x7fce5dbb1ec0, f=0x0) at /src/vpp/src/vlib/main.c:1502
#14 dispatch_process (vm=0x7fcfddded800 <vlib_global_main>, p=0x7fce5dbb1ec0, f=0x0, last_time_stamp=<optimized out>) at /src/vpp/src/vlib/main.c:1558
#15 0x0000000000000000 in ?? ()`

It seems dedicated upf session has been flushed, but traffic is still hitting upg proxy callbacks in separate thread

sergeymatov:
- If UPF session is already deleted by PFCP server, packets should not invoke proxy connection.
- if UPF session is in a deletion process (disabled, but not yet deleted), it will have no active rules. Prevent crash and don't connect proxy session in this case.

fixes #185 and #186

### DIFF
--- a/upf/upf_proxy.c
+++ b/upf/upf_proxy.c
@@ -364,6 +364,12 @@ proxy_start_connect_fn (const u32 * session_index)
 
   /* TODO: RACE: should not access the flow on the main thread */
   flow = pool_elt_at_index (fm->flows, ps->flow_index);
+  if (pool_is_free_index (gtm->sessions, flow->session_index))
+    {
+      ps->refcnt--;
+      ps->active_open_establishing = 0;
+      goto out;
+    }
   sx = pool_elt_at_index (gtm->sessions, flow->session_index);
   if (sx->generation != flow->generation)
     {
@@ -380,6 +386,12 @@ proxy_start_connect_fn (const u32 * session_index)
   ASSERT (flow_pdr_id (flow, FT_ORIGIN) != ~0);
   pdr = pfcp_get_pdr_by_id (active, flow_pdr_id (flow, FT_ORIGIN));
   ASSERT (pdr);
+  if (!pdr)
+    {
+      ps->refcnt--;
+      ps->active_open_establishing = 0;
+      goto out;
+    }
   far = pfcp_get_far_by_id (active, pdr->far_id);
 
   memset (a, 0, sizeof (*a));


upf/upf_proxy.c
```diff
@@ -364,6 +364,12 @@ proxy_start_connect_fn (const u32 * session_index)
 
   /* TODO: RACE: should not access the flow on the main thread */
   flow = pool_elt_at_index (fm->flows, ps->flow_index);
+  if (pool_is_free_index (gtm->sessions, flow->session_index))
```
ivan4th, line 367: This does not look exactly like a fix, isn't that more of a workaround? The idea which was mentioned in the comment above is that with a separate _single_ worker thread, we should only access the flowable from that worker thread. With multiple workers, things become harder as we have concurrent access to the flowtable from multiple workers. So perhaps we should add a comment that we still have a race here, and create a follow-up issue for proper multithreaded flowtable handling?

PR #192, opened 2021-11-29
[NAT] Fix NAT block calculation

sergeymatov:
Bihash search returns 0 if search succeeded. Port calculation loop should be going on if value for given combination of external address and port block has been found but wrong usage of returned value breaks this logic.

### DIFF
--- a/vpp-patches/0015-Controlled-NAT-function.patch
+++ b/vpp-patches/0015-Controlled-NAT-function.patch
@@ -1,4 +1,4 @@
-From 1366ac45f10a36d9fbf7e5abd449e40c16fd6c7f Mon Sep 17 00:00:00 2001
+From 2455718ddae87db22a2589b9e17dd967d0d26573 Mon Sep 17 00:00:00 2001
 From: Sergey Matov <sergey.matov@travelping.com>
 Date: Wed, 31 Mar 2021 15:19:02 +0400
 Subject: [PATCH] Controlled NAT function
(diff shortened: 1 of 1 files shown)


ivan4th (approved):
LGTM

mgumz:
@sergeymatov the commit message is a bit sparse on what this fix actually is doing and in which way the previous way of checking if an existing natip-portrange was already taken happened.

example given, the while smells like the algorithm is a doing a linear search over the availale natip-portranges until some free spot is found. could be wrong - but how the finding a free spot is working must be documented.

sergeymatov:
@mgumz 
port block allocation algorithm is not changed, it just was having a problem with validating if binding is already exist or not.
Original NAT binding calculation has been already discussed: we are looking for most free IP addr while tracking its usage, and for this IP addr we are trying to identify free port block.

mgumz:
so the `while` in this commit here is retry until a free binding is found in terms of collisions. 

still, looks like an opportunity for "endless-loop" … (until NAT_CONTROLLED_MAX_PORT) … so, whats the worst case scenario here?

sergeymatov:
no worst case scenario here. NAT_CONTROLLED_MAX_PORT stops port block selection, this means for most unused address no ports left. So, no binding will be created and PFCP establishment response will trigger an error.

mgumz:
well, 2 things:

1. the while loops unless `end` >= NAT_CONTROLLED_MAX_PORT … how often is this?
2. if no nat-binding on the picked nat-ip can be established, and this is communicated back to the SMC which communicates it back to TPOSS - what happens with the traffic? the traffic for that session still needs to be handled coz it wont stop. that is why its required to not stop describing the situation after the "pfcp establishment will trigger and error"

that said: the first packets for a given, yet unauthorized, framed-ip arrive which causes a the whole signalling dance involving NASREQ start comes first which just communicates the to-be-used nat-pool from tposs to upg. isnt then the pfcp-establishment done? iirc, the concrete NAT binding to ip/port-range happens when accounting really starts … what happens in this case with such a service-session in the upg?? how is the traffic from the UE handled in such a situation and how is it resolved?

sergeymatov:
1. if nat pool is out of capacity - all bindings are reserved
2. you are talking about the case of late binding that considered as non-supported in this case

PR #194, opened 2021-12-01
Add E2E tests for bypassing the proxy via higher-precendence PDRs

ivan4th:
This PR requires #190 which must be merged first, after which this one
needs to be updated (the single commit with the change must be rebased).


sergeymatov (approved):
LGTM

PR #196, opened 2021-12-02
[NAT] Fix NAT-ed for unknown protocol and port block calculation

sergeymatov:
In case on non-controlled NAT appropriate checks should be added
for unknown protocol translation.
Also, fixes issue with incorrect end port calculation for given
block size.

### DIFF
--- a/vpp-patches/0015-Controlled-NAT-function.patch
+++ b/vpp-patches/0015-Controlled-NAT-function.patch
@@ -1,4 +1,4 @@
-From 1366ac45f10a36d9fbf7e5abd449e40c16fd6c7f Mon Sep 17 00:00:00 2001
+From a84a30b9224d2d104e77bdbbf9177684889f3ecf Mon Sep 17 00:00:00 2001
 From: Sergey Matov <sergey.matov@travelping.com>
 Date: Wed, 31 Mar 2021 15:19:02 +0400
 Subject: [PATCH] Controlled NAT function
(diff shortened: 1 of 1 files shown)


ivan4th (approved):
LGTM

vpp-patches/0015-Controlled-NAT-function.patch
```diff
@@ -174,38 +174,61 @@ index 776efdf13..50b665f60 100644
 -  int i;
 +  ip4_address_t ext_addr;
    u8 is_sm = 0;
++  int i;
  
    switch (vec_len (sm->outside_fibs))
-@@ -839,17 +917,19 @@ nat44_ed_in2out_unknown_proto (snat_main_t * sm,
+     {
+@@ -839,21 +918,43 @@ nat44_ed_in2out_unknown_proto (snat_main_t * sm,
```
RoadRunnr, line 181: I'm not sure that this works for out use case.

If we do port block based NAT, then hitting a protocol where we can't control the port range (or a similar identifier) meant that we can not forward that packet.

I might work on the outwards (UE to SGi) path, but how would we ever map packets on the return path?
When doing by IP only, return packets might end up at the wrong UE -- and that is worse than not forwarding them.

sergeymatov: Good point. We might be wanted either to keep only one user allowed for unique combination of `nat_addr + dst_addr` or deny such traffic.  @RoadRunnr Do you think it's better to allow to do translations for single user or to drop anyway?

RoadRunnr: For a pool that is configured for 1:1 NAT, such sessions can be kept. For port based NAT, any such session must IMHO be dropped.

sergeymatov: It seems it's already supports only one user per unique `nat_addr + dst_addr` combination. For unknown protocol there are two lookups created - for out2in traffic and for in2out traffic. in2out traffic matches original src address. If traffic with different src addr but same `nat_addr + dst_addr` will appear, here
https://github.com/FDio/vpp/blob/stable/2101/src/plugins/nat/in2out_ed.c#L797
it won't be able to hit same session for already NATed traffic
and here
https://github.com/FDio/vpp/blob/stable/2101/src/plugins/nat/in2out_ed.c#L797
tuple for out2in is already existing, won't create a session and drop such traffic

RoadRunnr: PPTP can not work without NAT helper when multiple devices behind the same NAT IP try to reach the same PPTP server.

However, such a situation is more likely for ingress NAT, so it might not have been validate in out setups.

The NAT IP + Dst IP tuple is not a unique identifier without ports for unknown protocols. You would need to make sure that any internal IP has only one unique NAT IP + Dst IP tuple for outgoing connections. If that can be guaranteed, then its ok.

However, I doubt that this can be meet in the port range based NAT without helpers.

PR #198, opened 2021-12-02
[NAT] Fix NAT-ed for unknown protocol and port block calculation

sergeymatov:
In case on non-controlled NAT appropriate checks should be added
for unknown protocol translation.
Also, fixes issue with incorrect end port calculation for given
block size.


ivan4th (approved):
LGTM

PR #204, opened 2021-12-10
Restore changelog-config.json

ivan4th:
This file is used to generate release changelog and it was deleted by
accident during the split.


korroot (approved):
LGTM

PR #208, opened 2022-01-21
Add e2e test cases for the routing policy

ivan4th:
The proxied test case is disabled for now due to an issue with TCP
output subgraph (SYNs)

This PR includes changes from #194 which needs to be merged first.


sergeymatov (approved):
LGTM

PR #210, opened 2022-01-21
Check if input line is empty for upf policy CLI

sergeymatov:
If no input given, return and don't display anything.

### DIFF
--- a/upf/upf_cli.c
+++ b/upf/upf_cli.c
@@ -1503,28 +1503,29 @@ upf_policy_command_fn (vlib_main_t * vm,
   u8 *policy_id;
   u8 action = 0;
 
-  if (unformat_user (main_input, unformat_line_input, line_input))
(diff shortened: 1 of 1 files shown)


upf/upf_cli.c
```diff
@@ -1503,6 +1503,9 @@ upf_policy_command_fn (vlib_main_t * vm,
   u8 *policy_id;
   u8 action = 0;
 
+  if (!unformat_user (main_input, unformat_line_input, line_input))
+    return 0;
+
```
ivan4th, line 1508: Why invoke `unformat_user()` twice?
Other place the input is handled likewise, for reference: https://github.com/travelping/upg-vpp/blob/master/upf/upf_cli.c#L60-L63

sergeymatov: good point, fixed

ivan4th (approved):
LGTM

PR #211, opened 2022-01-26
Fix forwarding policy handling for the proxy output

ivan4th:
TODO:
- [x] merge https://github.com/travelping/vpp-base/pull/2 and update this PR

This directs all of the TCP connection packets for both active and
passive proxy connections through the UPG subgraph. Care is taken not
do traffic accounting twice on the same packets. This also applies to
RSTs that are sent by the proxy to terminate a particular connection
upon an error. TCP frames not related to a particular connection, such
as RSTs sent in response to late frames after the connection is
already gone, still don't go through the subgraph at the moment, but
fixing that part would be more problematic as it would require flow
lookup without having SEID.

Newer VPP versions (21.10) does not have separate half-open and normal
connection pools and uses `tcp[46]-output` nodes for all of the
connection-related packets, which is followed by the node that can be
specified per-connection, so there will be a possibility to simplify
the approach. At the moment, only the part that adds next node index /
opaque value is backported (travelping/vpp-base#2), and the opaque
value storing the flow id is obtained by looking up the TCP
connection.

### DIFF
--- a/upf/upf_proxy_accept.c
+++ b/upf/upf_proxy_accept.c
@@ -145,9 +145,9 @@ proxy_session_stream_accept_notify (transport_connection_t * tc, u32 flow_id)
 /**
  * Lookup transport connection
  */
-static tcp_connection_t *
+tcp_connection_t *
 upf_tcp_lookup_connection (u32 fib_index, vlib_buffer_t * b, u8 thread_index,
-			   u8 is_ip4)
+			   u8 is_ip4, u8 is_reverse)
 {
   tcp_header_t *tcp;
   transport_connection_t *tconn;
@@ -158,13 +158,23 @@ upf_tcp_lookup_connection (u32 fib_index, vlib_buffer_t * b, u8 thread_index,
       ip4_header_t *ip4;
       ip4 = vlib_buffer_get_current (b);
       tcp = ip4_next_header (ip4);
-      tconn = session_lookup_connection_wt4 (fib_index,
-					     &ip4->dst_address,
-					     &ip4->src_address,
-					     tcp->dst_port,
-					     tcp->src_port,
-					     TRANSPORT_PROTO_TCP,
-					     thread_index, &is_filtered);
+      if (is_reverse)
+	tconn = session_lookup_connection_wt4 (fib_index,
+					       &ip4->src_address,
+					       &ip4->dst_address,
+					       tcp->src_port,
+					       tcp->dst_port,
+					       TRANSPORT_PROTO_TCP,
+					       thread_index, &is_filtered);
+      else
+	tconn = session_lookup_connection_wt4 (fib_index,
+					       &ip4->dst_address,
+					       &ip4->src_address,
+					       tcp->dst_port,
+					       tcp->src_port,
+					       TRANSPORT_PROTO_TCP,
+					       thread_index, &is_filtered);
+
       tc = tcp_get_connection_from_transport (tconn);
       /* ASSERT (tcp_lookup_is_valid (tc, b, tcp)); */
     }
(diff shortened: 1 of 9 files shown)


upf/upf_proxy_output.c
```diff
@@ -149,9 +153,99 @@ upf_proxy_output (vlib_main_t * vm, vlib_node_runtime_t * node,
...
+	       * actual direction, which is done after we get the
+	       * flow.
+	       */
+	      direction = is_ip4 ? ip4_packet_is_reverse (ip4) :
+		ip6_packet_is_reverse (ip6);
+	      fib_idx = vlib_buffer_get_ip_fib_index (b, is_ip4);
+	      /*
+	       * We need the flow to understand the direciton of the
+	       * connection.  But also we need to get the connection
+	       * to get the flow.  To break out of this chcken-and-egg
```
sergeymatov, line 201: chicken

ivan4th: thanks, fixed, along with another typo nearby (`direciton`)

sergeymatov (approved):

upf/upf_proxy_accept.c
```diff
@@ -159,9 +159,13 @@ upf_tcp_lookup_connection (u32 fib_index, vlib_buffer_t * b, u8 thread_index,
...
       tcp = ip4_next_header (ip4);
       tconn = session_lookup_connection_wt4 (fib_index,
+					     is_reverse ? &ip4->src_address :
 					     &ip4->dst_address,
+					     is_reverse ? &ip4->dst_address :
 					     &ip4->src_address,
+					     is_reverse ? tcp->src_port :
 					     tcp->dst_port,
+					     is_reverse ? tcp->dst_port :
 					     tcp->src_port,
```
RoadRunnr, line 169: The excessive ?: looks ugly. Wouldn't it be better to have a pain `if ... else` only?

ivan4th: Well, that adds another `if` nesting level which is also a bit dubious, esp. given that it's repeated for IPv4 and IPv6, and also that looked a bit cleaner before `indent` had split the lines ;) But ok, will redo with `if`s

RoadRunnr (commented):
What I don't get is why the NO_COUNT logic is needed. How would a packet ever traverse tcp-output multiple times?

upf/upf_proxy_accept.c
RoadRunnr: maybe use a wrapper around session_lookup_connection_wt4?

ivan4th:
@RoadRunnr same packet can't go through that path. But consider this situation:
the data goes in the uplink direction through the proxy (non-stitched). Along the lines of:
```
ethernet-input -> ip4-input -> ip4-lookup -> lookup-ip4-src -> upf-ip4-session-dpo ->
  upf-ip4-flow-process -> upf-ip4-proxy-input -> tcp4-input-nolookup -> tcp4-rcv-process
```
Here, in `upf-ip4-proxy-input`, it is accounted for by calling `process_urrs()`. Then the proxy sends this packet further. Before this PR, that looked like this:
```
tcp4-output -> ip4-lookup -> ip4-rewrite -> host-sgi-output -> host-sgi-tx
```
Whereas after it:
```
tcp4-output -> upf-ip4-proxy-server-no-count-output -> upf-ip4-forward ->
ip4-lookup -> ip4-rewrite -> host-sgi-output -> host-sgi-tx
```

Here, `upf-ip4-forward` node will take care of handling the forwarding parameters, incl. forwarding policy; yet, by default, it also invokes `process_urrs()`. But, as this particular piece of data, even though while being part of another frame, is already accounted for before, we will have it counted twice unless we set a hint in `upf-ip4-proxy-server-no-count-output` so as to avoid this situation.

RoadRunnr:
@ivan4th took me some time to correct my mental picture. I get now why the output node wasn't in the chain before.

Still, naming it "NO_COUNT" seems misleading. Especially since it skips QER handling as well.

What do you think about replacing `no_count` with `far_only` ???

ivan4th:
That makes sense, will fix

upf/upf_proxy_accept.c
ivan4th: In the end, another nested `if` didn't look that bad

ivan4th:
Replaced "no count" with "far only"

korroot (approved):
LGTM

PR #213, opened 2022-01-26
Forwarding Policy protocol selection

sergeymatov:
Single forwarding policy should handle routing paths for both IPv4 and IPv6.
Appropriate route path lookup is added in favor of traffic protocol.

### DIFF
--- a/test/e2e/upg_e2e.go
+++ b/test/e2e/upg_e2e.go
@@ -1362,18 +1362,21 @@ func describeGTPProxy(title string, ipMode framework.UPGIPMode) {
 func describeRoutingPolicy(f *framework.Framework) {
 	ginkgo.Describe("routing policy", func() {
 		var altServerIP *net.IPNet
+		var ipTable uint32
 		ginkgo.BeforeEach(func() {
+			f.VPP.Ctl("ip table add 201")
+			f.VPP.Ctl("ip6 table add 301")
+			f.VPP.Ctl("upf policy add id altIP via ip4-lookup-in-table 201 via ip6-lookup-in-table 301")
+
 			if f.IPMode == framework.UPGIPModeV4 {
 				altServerIP = framework.MustParseIPNet("192.168.99.3/32")
-				f.VPP.Ctl("ip table add 201")
-				f.VPP.Ctl("upf policy add id altIP via ip4-lookup-in-table 201")
+				ipTable = 201
 			} else {
 				altServerIP = framework.MustParseIPNet("2001:db8:aa::3/128")
-				f.VPP.Ctl("ip6 table add 201")
-				f.VPP.Ctl("upf policy add id altIP via ip6-lookup-in-table 201")
+				ipTable = 301
 			}
 			f.AddCustomServerIP(altServerIP)
-			f.VPP.Ctl("ip route add %s table 201 via %s host-sgi0", altServerIP, f.ServerIP())
+			f.VPP.Ctl("ip route add %s table %d via %s host-sgi0", altServerIP, ipTable, f.ServerIP())
 		})
 
 		verify := func(sessionCfg framework.SessionConfig) {
--- a/upf/upf_forward.c
+++ b/upf/upf_forward.c
@@ -198,28 +198,6 @@ upf_forward (vlib_main_t * vm, vlib_node_runtime_t * node,
 		      goto trace;
 		    }
 		}
-	      else if (far->forward.flags & FAR_F_FORWARDING_POLICY)
-		{
-		  if (is_ip4)
-		    b->flags &= ~(VNET_BUFFER_F_OFFLOAD_TCP_CKSUM |
-				  VNET_BUFFER_F_OFFLOAD_UDP_CKSUM |
-				  VNET_BUFFER_F_OFFLOAD_IP_CKSUM);
-		  else
-		    b->flags &= ~(VNET_BUFFER_F_OFFLOAD_TCP_CKSUM |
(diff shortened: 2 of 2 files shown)


upf/upf_forward.c
```diff
@@ -213,8 +214,16 @@ upf_forward (vlib_main_t * vm, vlib_node_runtime_t * node,
...
+		  vec_foreach (rpath, fp_entry->rpaths)
+		  {
+		    if (is_ip4 && rpath->frp_proto == DPO_PROTO_IP4)
+		      break;
+		    if (!is_ip4 && rpath->frp_proto == DPO_PROTO_IP6)
+		      break;
+		  }
 		  vnet_buffer (b)->sw_if_index[VLIB_TX] =
-		    fp_entry->rpaths->frp_fib_index;
+		    rpath->frp_fib_index;
```
ivan4th, line 226: Why it is fib_index, but assigned to `sw_if_index` value? From what I understand, interfaces have their own indexing that does not directly correspond to FIB indices

upf/upf_forward.c
```diff
@@ -213,8 +214,16 @@ upf_forward (vlib_main_t * vm, vlib_node_runtime_t * node,
 				       far->forward.fp_pool_index);
 		  vnet_buffer (b)->ip.adj_index[VLIB_TX] =
 		    fp_entry->dpo.dpoi_index;
+
+		  vec_foreach (rpath, fp_entry->rpaths)
+		  {
+		    if (is_ip4 && rpath->frp_proto == DPO_PROTO_IP4)
```
ivan4th, line 220: maybe `rpath->frp_proto == (is_ip4 ? DPO_PROTO_IP4 : DPO_PROTO_IP6)`?

test/e2e/upg_e2e.go
```diff
@@ -1362,18 +1362,21 @@ func describeGTPProxy(title string, ipMode framework.UPGIPMode) {
 func describeRoutingPolicy(f *framework.Framework) {
 	ginkgo.Describe("routing policy", func() {
 		var altServerIP *net.IPNet
+		var ipTable string
```
ivan4th, line 1365: Why is it string not int/uint?

ivan4th (changes requested):

upf/upf_forward.c
ivan4th: Ok I see:
https://github.com/FDio/vpp/blob/a424dd1/src/vnet/ip/ip4_forward.c#L80-L86

upf/upf_forward.c
sergeymatov: good point. done

test/e2e/upg_e2e.go
sergeymatov: redone for u32

ivan4th (approved):
LGTM

korroot (approved):
LGTM

upf/upf_forward.c
```diff
@@ -211,10 +212,15 @@ upf_forward (vlib_main_t * vm, vlib_node_runtime_t * node,
...
 				       far->forward.fp_pool_index);
-		  vnet_buffer (b)->ip.adj_index[VLIB_TX] =
-		    fp_entry->dpo.dpoi_index;
+
+		  vec_foreach (rpath, fp_entry->rpaths)
+		  {
+		    if (rpath->frp_proto ==
+			(is_ip4 ? DPO_PROTO_IP4 : DPO_PROTO_IP6))
+		      break;
+		  }
```
RoadRunnr, line 221: what happens when there is entry that matches the current protocol version?

If nothing matches, then rpath will end up pointing to a element after the end of vector. That is an invalid entry and the index value in there is invalid.

RoadRunnr (changes requested):

upf/upf_forward.c
sergeymatov: Very good point, thanks.
If policy is underconfigured it will lead to crash here.
Fixed this to use normal forwarding based on FAR's NWI.

upf/upf_forward.c
```diff
@@ -211,10 +212,45 @@ upf_forward (vlib_main_t * vm, vlib_node_runtime_t * node,
...
+
+		  if (!rpath)
```
RoadRunnr, line 223: that test is not going to work. rpath will not be `NULL` if the vec_foreach reached the end of vector.

This might:
```suggestion
		  if (var < vec_end (rpath))
```

sergeymatov: added a check for `rpath < vec_end (rpaths)`

RoadRunnr (approved):
LGTM

upf/upf_forward.c
```diff
@@ -198,28 +198,6 @@ upf_forward (vlib_main_t * vm, vlib_node_runtime_t * node,
...
-				  VNET_BUFFER_F_OFFLOAD_UDP_CKSUM |
-				  VNET_BUFFER_F_OFFLOAD_IP_CKSUM);
-		  else
-		    b->flags &= ~(VNET_BUFFER_F_OFFLOAD_TCP_CKSUM |
-				  VNET_BUFFER_F_OFFLOAD_UDP_CKSUM);
-		  /* Getting dpio_index */
-		  fp_entry =
-		    pool_elt_at_index (gtm->upf_forwarding_policies,
-				       far->forward.fp_pool_index);
-		  vnet_buffer (b)->ip.adj_index[VLIB_TX] =
```
RoadRunnr, line 214: is this still needed? Because I might have accidently removed it!

PR #219, opened 2022-02-22
Initial IPFIX implementation

ivan4th:
- [x] Initial implementation
- [x] IMSI support
- [x] E2E test
- [x] IPv6 support
- [x] multiple templates (FAR-based, using the vendor-specific IE)
- [x] NWI-based template selection
- [x] Multiple exporters
- [x] NAT fields
- [x] test with the proxy (IPFIX policy selection based on PDRs with ADF)
- [x] verify User ID IE in the Scapy-based tests
- [x] sort out the branches

TODO in a follow-up:
- other missing fields (VRF IDs)
- make reporting interval configurable
- fix format_tbcd (non-numeric chars)

IMSI in the IPFIX packets is presently encoded using the format described in TS 29.274 Section 8.3.

[go-ipfix](https://github.com/vmware/go-ipfix) is being used for E2E testing, fixes:
https://github.com/vmware/go-ipfix/pull/276
https://github.com/vmware/go-ipfix/pull/277


upf/upf_ipfix.c
```diff
@@ -0,0 +1,958 @@
...
+{
+  u16 start = offset;
+  upf_main_t *gtm = &upf_main;
+  upf_session_t *sx = pool_elt_at_index (gtm->sessions, f->session_index);
+
+  /* mobileIMSI */
+  /* the value can't be 255 or more due to limitations of pfcp.c, */
+  /* thus we use simpler encoding */
+  to_b->data[offset] = sx->imsi_len;
+  clib_memcpy_fast (to_b->data + offset + 1, sx->imsi, sx->imsi_len);
```
miguelbf-alb, line 474: I gave this a whirl out of curiosity for the functionality and while pretty much everything works out of the box the imsi ipfix field is not decoded by wireshark or any ipfix collector I've tried. As far as I could see the mobileImsi is sent in the [ipfix payload as a string](https://www.iana.org/assignments/ipfix/ipfix.xhtml) so, considering the imsi has a length of 15 codepoints the provided buffer should be 16 (imsi + null terminator).  `sx->imsi` is currently stored as 8 bytes though.
I guess here you need to format the mobile identify and allocate a 16 buffer for this to work:

```c
diff --git a/upf/upf_ipfix.c b/upf/upf_ipfix.c
index e493271..4756036 100644
--- a/upf/upf_ipfix.c
+++ b/upf/upf_ipfix.c
@@ -470,9 +470,15 @@ upf_ipfix_common_add (vlib_buffer_t * to_b, flow_entry_t * f, flow_direction_t d
   /* mobileIMSI */
   /* the value can't be 255 or more due to limitations of pfcp.c, */
   /* thus we use simpler encoding */
-  to_b->data[offset] = sx->imsi_len;
-  clib_memcpy_fast (to_b->data + offset + 1, sx->imsi, sx->imsi_len);
-  offset += sx->imsi_len + 1;
+
+  // note: imsi string is 15 code points and is currently stored as an 8 bit array (u8* -> sizeof(sx->imsi)=8).
+  // the ipfix rfc state that the imsi is encoded as a string (unicode uses 1 byte per codepoint for the first 128 codepoints). Hence when transfering the imsi in the ipfix payload
+  // we need to pre-allocate a buffer of size 16 before calling format_mobile_identity (15 codepoints + null terminator). This is equivalent to  2 * sx->imsi_len + 1.
+
+  to_b->data[offset] = sx->imsi_len * 2;
+  u8 *imsi = format (NULL, "%U", format_mobile_identity, sx->imsi, sx->imsi_len);
+  clib_memcpy_fast (to_b->data + offset + 1, imsi, 2 * sx->imsi_len);
+  offset += 2 * sx->imsi_len + 1;
```

Snippet quality might not be the best one ever but works as a PoC for me :)

miguelbf-alb: Btw I also got a segmentation fault on the vpp startup, gdb pointed out at the inability of allocating memory for the flow table. I had to reduce its size for this to work:

```
diff --git a/upf/flowtable.h b/upf/flowtable.h
index e535d87..467242d 100644
--- a/upf/flowtable.h
+++ b/upf/flowtable.h
@@ -207,7 +207,7 @@ typedef struct
  * hashtable is configured to alloc (NUM_BUCKETS * CLIB_CACHE_LINE_BYTES) Bytes
  * with (flow_count / (BIHASH_KVP_PER_PAGE / 2)) Buckets
  */
-#define FM_POOL_COUNT_LOG2 22
+#define FM_POOL_COUNT_LOG2 15
 #define FM_POOL_COUNT (1 << FM_POOL_COUNT_LOG2)
 #define FM_NUM_BUCKETS (1 << (FM_POOL_COUNT_LOG2 - (BIHASH_KVP_PER_PAGE / 2)))
 #define FM_MEMORY_SIZE (FM_NUM_BUCKETS * CLIB_CACHE_LINE_BYTES * 6)
```

Doesn't seem to happen on master (nor with the base branch - 22.02 so likely has to do with the ipfix addition.

ivan4th: Sorry for the late reply.
Concerning `mobileIMSI` field: I had an impression that it must be stored using TBCD encoding, but will recheck
Concerning the segfault (actually SIGABRT I think due to heap exhaustion): you may try increasing the heap size, e.g. by setting `heapsize 3G` in startup.conf

upf/pfcp.c
```diff
@@ -2496,6 +2496,23 @@ format_outer_header_creation (u8 * s, va_list * args)
   return s;
 }
 
+u8 *format_mobile_identity (u8 * s, va_list * args)
```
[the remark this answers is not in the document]
ivan4th: Fixed

upf/pfcp.c
```diff
@@ -2496,6 +2496,23 @@ format_outer_header_creation (u8 * s, va_list * args)
...
+{
+  u8 *bytes = va_arg (*args, u8 *);
+  int n_bytes = va_arg (*args, int);
+  uword i;
+
+  for (i = 0; i < n_bytes; i++)
+    {
+      if (bytes[i] & 0xf0 == 0xf0 && i == n_bytes - 1)
```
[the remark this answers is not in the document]
ivan4th: done

upf/pfcp.h
```diff
@@ -1003,8 +1003,8 @@ typedef u8 pfcp_suggested_buffering_packets_count_t;
 #define PFCP_IE_USER_ID					141
 typedef struct
 {
-#define USER_ID_IMEI					BIT(0)
-#define USER_ID_IMSI					BIT(1)
+#define USER_ID_IMSI					BIT(0)
+#define USER_ID_IMEI					BIT(1)
 #define USER_ID_MSISDN					BIT(2)
 #define USER_ID_NAI					BIT(4)
```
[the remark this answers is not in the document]
ivan4th: Fixed (also verified by the scapy-based test)

upf/upf_flow_node.c
```diff
@@ -244,22 +249,11 @@ upf_flow_process (vlib_main_t * vm, vlib_node_runtime_t * node,
...
-	   * avoid scheduling flows "in the past"
-	   */
-	  flow_update_lifetime (flow0, p0, is_ip4);
-	  flow_update_lifetime (flow1, p1, is_ip4);
-
-	  /* flow statistics */
-	  flow0->stats[is_reverse0].pkts++;
-	  flow0->stats[is_reverse0].bytes += b0->current_length;
-	  flow1->stats[is_reverse1].pkts++;
-	  flow1->stats[is_reverse1].bytes += b1->current_length;
```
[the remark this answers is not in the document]
ivan4th: Well, I'm not sure inline funcs will be much worse than repeated code / macros (that is: perhaps the same code will be generated in the end), although perhaps I don't know enough of clang internals to be sure about that. If further testing will reveal performance degradation, I'll take a closer look at this place; although so far the main bottleneck in UPG appears to be insufficient cache locality

upf/flowtable.c
```diff
@@ -639,6 +647,45 @@ VLIB_CLI_COMMAND (upf_show_flow_timeout_command, static) =
...
+      vlib_worker_thread_barrier_release (vm);
+    }
+
+  return 0;
+}
+
+/* *INDENT-OFF* */
+VLIB_REGISTER_NODE (flowtable_process_node) = {
+  .function = flowtable_process,
+  .type = VLIB_NODE_TYPE_PROCESS,
```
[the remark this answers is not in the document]
ivan4th: Just replacing `VLIB_NODE_TYPE_PROCESS` with `VLIB_NODE_TYPE_PRE_INPUT` causes the code to crash, so it needs a bit more adjustments to do it that way. Perhaps could do it in a follow-up.
This node is disabled by default and is only used in testing situations with low traffic (causing some remaining flows not to expire) and interrupt mode in use.

upf/upf.api
```diff
@@ -223,15 +235,19 @@ autoreply define upf_nwi_add_del {
...
 };
 
 define upf_nwi_details {
   u32 client_index;
   u32 context;
-  u32 ip4_fib_table;
-  u32 ip6_fib_table;
+  u32 ip4_table_id;
+  u32 ip6_table_id;
+  u8 ipfix_policy[32];
```
RoadRunnr, line 249: limiting the policy name to 31 characters is a bit strange.

upf/upf.h
```diff
@@ -785,6 +792,9 @@ typedef struct
 
   f64 unix_time_start;
 
+  u8 imsi_len;
+  u8 imsi[8];
```
RoadRunnr, line 796: additional fields from IMEI and MSISDN would be needed as well. Wouldn't it be simpler to put a reference to or a copy of the PFCP User Id IE here?

upf/upf_classify.c
```diff
@@ -278,6 +278,26 @@ upf_acl_classify_forward (vlib_main_t * vm, u32 teid, flow_entry_t * flow,
 	      *pdr_idx = acl->pdr_idx;
 
 	    far = pfcp_get_far_by_id (active, pdr->far_id);
+	    flow->ipfix_context_index = is_ip4 ?
+	      far->ipfix_context_index_ip4 : far->ipfix_context_index_ip6;
+	    /*
+	     * If IPFIX policy is not set in the FAR, use the value
+	     * from the session
+	     */
+	    if (flow->ipfix_context_index == (u32) ~ 0)
```
RoadRunnr, line 287: If you unconditionally use the sessions IPFIX settings, how can a FAR then disable the inclusion of its flow in the export?
If should IMHO be possible to specify a empty IPFIX policy at the FAR level and use that to override any session level or default IPFIX export rules.

upf/upf_ipfix.c
```diff
@@ -0,0 +1,693 @@
...
+				 cidx, &context->template_id,
+				 context->is_ip4, false);
+  if (rv)
+    clib_warning ("couldn't remove IPFIX report, perhaps "
+		  "the exporter has been deleted?");
+
+  vec_free (context->buffers_per_worker);
+  vec_free (context->frames_per_worker);
+  vec_free (context->next_record_offset_per_worker);
+  pool_put (fm->contexts, context);
```
RoadRunnr, line 606: pool_put is not thread safe. The unref method can be called from multiple threads, when this happens, the put will corrupt the pool.

upf/upf_ipfix.c
```diff
@@ -0,0 +1,693 @@
...
+void
+upf_unref_ipfix_context_by_index (u32 cidx)
+{
+  int rv;
+  upf_ipfix_main_t *fm = &upf_ipfix_main;
+  clib_bihash_kv_24_8_t kv;
+  upf_ipfix_protocol_context_t *context;
+
+  context = pool_elt_at_index (fm->contexts, cidx);
+  if (--context->ref_count)
```
RoadRunnr, line 588: upf_unref_ipfix_context_by_index can could be invoked by multiple threads for the same flow. So updating the ref_count should use a atomic function (probably in other places as well).

upf/flowtable.h
```diff
@@ -88,7 +88,11 @@ typedef struct
 typedef struct
 {
   u32 pkts;
+  u32 pkts_unreported;
   u64 bytes;
+  u64 bytes_unreported;
+  u64 l4_bytes;
+  u64 l4_bytes_unreported;
```
RoadRunnr, line 95: I really can't see why `l4_bytes` and `l4_bytes_unreported` is needed. It seems that nothing is using `l4_bytes` and both fields will have the same value.
THe same might apply to pkts and bytes.

RoadRunnr (changes requested):
I'm not convinced that the interaction of flow table hooks and the IPFIX contexts is even remotely thread safe. To me it looks as if multiple threads would trample all over the shared data structures and corrupt them.

test/e2e/ipfix_handler.go
```diff
@@ -0,0 +1,171 @@
+package exttest
```
sergeymatov, line 1: copyrights probably needed for new files

upf/flowtable.h
ivan4th: That actually looks like a bug b/c at some point, `l4_bytes_unreported` were being cleared for the flow when the report was being sent, but somehow I lost that part and somehow the test did not detect the issue. I will recheck the code, thanks. Plain `l4_bytes` is for the "total" counter starting from the beginning of the flow, presently the fields that are using it aren't added to the existing templates but we might easily need them.

upf/upf_ipfix.c
```diff
@@ -0,0 +1,693 @@
...
+void upf_ipfix_flow_update_hook (flow_entry_t * f, flow_direction_t direction, u32 now)
+{
+  upf_ipfix_main_t *fm = &upf_ipfix_main;
+  vlib_main_t *vm = fm->vlib_main;
+
+  if (fm->disabled || f->ipfix_context_index == (u32) ~0)
+      return;
+
+  if (fm->active_timer == 0
+      || (now > flow_last_exported(f, direction) + fm->active_timer))
```
sergeymatov, line 477: Thinking about NAT problem, this might be the correct place to check if `flow_last_exported` is positive before actual exporting. wdyt @ivan4th ?

sergeymatov (changes requested):

upf/flowtable.h
ivan4th: Rechecked it and there's no bug except that the tests needed a bit of adjustment to check it's all working correctly (will push the changes).
Note that for `initiatorPackets` / `responderPackets` / `initiatorOctets` / `responderOctets` we use the macro `IPFIX_VALUE_DELTA_U64`
https://github.com/travelping/upg-vpp/blob/feature/ipfix/upf/upf_ipfix_templates.h#L127
Which does reset the corresponding flow field to zero after copying the value to the IPFIX report
https://github.com/travelping/upg-vpp/blob/feature/ipfix/upf/upf_ipfix_templates.h#L68
Thus, `..._unreported` here stands for the delta fields for which the values are reset to zero during each IPFIX report (that is, "unreported" in the sense "the amount which is not reported yet")

upf/upf_ipfix.c
```diff
@@ -0,0 +1,693 @@
...
+{
+  upf_ipfix_policy_t policy;
+  u32 name_len = vec_len (name);
+
+  if (ok)
+    *ok = false;
+
+  for (policy = UPF_IPFIX_POLICY_NONE; policy < UPF_IPFIX_N_POLICIES; policy++)
+    {
+      u32 l = strlen(upf_ipfix_templates[policy].name);
```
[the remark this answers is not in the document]
ivan4th: Problem is, `upf_ipfix_templates[policy].name` is not a vector but rather a string (specified as a string literal in the code) and thus `vec_is_equal` will not work for it

upf/upf.api
ivan4th: Extended it to 64 bytes, perhaps that's should be enough for now :)
Presently, the set of policies is fixed (only `none`, `dest` and `default`)

upf/upf.h
ivan4th: Fixed, now it has PFCP User ID structure here

upf/upf_classify.c
ivan4th: Fixed, now it's possible to specify an empty string as IPFIX policy in the FAR to override the NWI setting.

upf/upf_ipfix.c
ivan4th: Added spinlock

upf/upf_ipfix.c
ivan4th: Switched to atomic func

test/e2e/ipfix_handler.go
ivan4th: Fixed

upf/upf_ipfix.c
```diff
@@ -0,0 +1,693 @@
...
+  b0->current_length = offset;
+
+  context->next_record_offset_per_worker[my_cpu_number] = offset;
+  /* Time to flush the buffer? */
+  /* TODO uncomment! also: force upon removal */
+  /* if (offset + context->rec_size > exp->path_mtu) */
+  upf_ipfix_export_send (vm, b0, context, now);
+}
+
+#define vec_neg_search(v,E)			\
```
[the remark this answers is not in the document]
ivan4th: Removed, it's unused indeed

upf/upf_ipfix.c
ivan4th: Well, in this case, it would never be exported till the flow ends ;) Added setting `flow_last_exported` to `now` upon  initialization of a particular flow

upf/upf_api.c
```diff
@@ -485,26 +488,64 @@ vl_api_upf_nwi_add_del_t_handler (vl_api_upf_nwi_add_del_t * mp)
...
+      goto out;
+    }
 
   vec_validate (nwi_name, mp->nwi_len - 1);
   memcpy (nwi_name, mp->nwi, mp->nwi_len);
   ip4_table_id = clib_net_to_host_u32 (mp->ip4_table_id);
   ip6_table_id = clib_net_to_host_u32 (mp->ip6_table_id);
-  if (!ip4_table_id && !ip6_table_id)
+
+  clib_warning ("(1) ip4_table_id %u, ip6_table_id %u", ip4_table_id,
```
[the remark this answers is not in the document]
ivan4th: Removed

sergeymatov (approved):
Overal - LGTM

upf/flowtable.h
RoadRunnr: Sure, that applies to l4_bytes_unreported and pkts_unreported. There is no user for l4_bytes and bytes_unreported. I wonder if that is an oversight?

upf/upf_pfcp_api.c
```diff
@@ -2570,6 +2608,9 @@ handle_session_establishment_request (pfcp_msg_t * msg,
       pending->inactivity_timer.handle = ~0;
     }
 
+  if (ISSET_BIT (req->grp.fields, SESSION_ESTABLISHMENT_REQUEST_USER_ID))
+    memcpy (&sess->user_id, &req->user_id, sizeof (pfcp_user_id_t));
```
RoadRunnr, line 2612: That doesn't work for  the `nai` field. We don't use it at the moment, but to avoid problems in the future it would be better initialize it to zero in the copy.
If you wanted to copy it as well, then it needs to freed together with the session

upf/upf_ipfix_templates.c
```diff
@@ -0,0 +1,172 @@
...
+  IPFIX_FIELD_POST_NAT_SOURCE_IPV4_ADDRESS(F)	\
+  IPFIX_FIELD_POST_NAPT_SOURCE_TRANSPORT_PORT(F)
+
+#define IPFIX_TEMPLATE_DEFAULT_IPV6(F)		\
+  IPFIX_FIELD_SOURCE_IPV6_ADDRESS(F)		\
+  IPFIX_FIELD_DESTINATION_IPV6_ADDRESS(F)	\
+  IPFIX_FIELD_PROTOCOL_IDENTIFIER(F)
+
+#define IPFIX_TEMPLATE_DEFAULT_COMMON(F)	\
+  IPFIX_FIELD_MOBILE_IMSI(F)			\
```
RoadRunnr, line 46: what about IMSI and IMEI ?

RoadRunnr (approved):
The PR is too large for an in depth review. I have a few comments, but considering the size of this beast, those could be addressed in a follow up PR.

PR #222, opened 2022-02-22
[NAT] Fix Created NAT Binding Port range calculation

sergeymatov:
Since port range should be exactly block_size length and start port
is included into it, calculation of end port was incorrect.
Actual calculation for binding is correct, however value for report to
CP-F should be fixed.

### DIFF
--- a/upf/upf_pfcp_api.c
+++ b/upf/upf_pfcp_api.c
@@ -804,7 +804,7 @@ upf_alloc_and_assign_nat_binding (upf_nat_pool_t * np, upf_nat_addr_t * addr,
 			    np->port_block_size);
   if (port_start)
     {
-      port_end = port_start + np->port_block_size;
+      port_end = port_start + np->port_block_size - 1;
       addr->used_blocks += 1;
       sx->nat_addr = addr;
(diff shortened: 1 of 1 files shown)


sergeymatov:
> "Actual calculation for binding is correct, however value for report to CP-F should be fixed."
> 
> Is this commit fixing the issue or "should it be fixed by another commit"?

Should be fixed by this commit. Statement was made in compare to how it's actually works for NAT Binding

mgumz:
ah ok.

so, the report of the actual used nat-port was wrong _or_ was the reported nat-port-range incorrect?

sergeymatov:
Port range reported from UP-F to CP-F was incorrect, but actual used nat-port was correct

mgumz:
If the Port-Range was incorrect … then other functions relying on that report are also affected. I just bumped upg-vpp to 1.2.3 in an integration cluster in the assumption that the incorrect port-usage was the only issues.

Ok, will you provide a backport to the 1.2.x branch for this commit as well?

sergeymatov:
@mgumz sure, right after we will merge this one

korroot (approved):
LGTM

PR #231, opened 2022-04-04
Remove VTIME out of measurement methods

sergeymatov:
TS 29.244 Clause 5.2.2.2: VTIME should be invoked regardlessly to chosen measurement methods.
This fix moves Quota Validity timer out of measurement period checking.


ivan4th (approved):
LGTM

PR #233, opened 2022-04-08
Token bucket PFCP Session Report limiter

sergeymatov:
Generic rate limiter based on vnet/policer

### DIFF
--- a/upf/upf.c
+++ b/upf/upf.c
@@ -207,6 +207,34 @@ vnet_upf_nat_pool_add_del (u8 * nwi_name, ip4_address_t start_addr,
   return 0;
 }
 
+void
+upf_pfcp_policer_config_init (upf_main_t * gtm)
+{
+  qos_pol_cfg_params_st *cfg = &pfcp_rate_cfg_main;
+
+  cfg->rate_type = QOS_RATE_PPS;
+  cfg->rnd_type = QOS_ROUND_TO_CLOSEST;
+  cfg->rfc = QOS_POLICER_TYPE_1R2C;
+  cfg->color_aware = 0;
+  cfg->conform_action.action_type = QOS_ACTION_TRANSMIT;
+  cfg->exceed_action.action_type = QOS_ACTION_DROP;
+  cfg->violate_action.action_type = QOS_ACTION_DROP;
+  cfg->rb.pps.cir_pps = 1 << 20;
+  cfg->rb.pps.cb_ms = 1000;
+}
+
+void
+upf_pfcp_policers_recalculate (qos_pol_cfg_params_st * cfg)
+{
+  upf_main_t *gtm = &upf_main;
+  policer_t *p;
+
+  pool_foreach (p, gtm->pfcp_policers)
+  {
+    pol_logical_2_physical (cfg, p);
+  }
+}
+
 int
 vnet_upf_upip_add_del (ip4_address_t * ip4, ip6_address_t * ip6,
 		       u8 * name, u8 intf, u32 teid, u32 mask, u8 add)
@@ -581,6 +609,8 @@ upf_init (vlib_main_t * vm)
   error = flowtable_init (vm);
   if (error)
     return error;
+
+  upf_pfcp_policer_config_init (sm);
 
   return pfcp_server_main_init (vm);
 }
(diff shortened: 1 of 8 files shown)


upf/upf.c
```diff
@@ -195,6 +195,24 @@ vnet_upf_nat_pool_add_del (u8 * nwi_name, ip4_address_t start_addr,
   return 0;
 }
 
+void
+upf_pfcp_policer_init (upf_main_t * vm)
```
ivan4th, line 199: `upf_main_t *` vars are (historically) called `gtm` or sometimes `sm`, but never `vm` which is used for `vlib_main_t *`. Perhaps someday we should rename them to `um`, but at least let's not make the naming even more diverse

upf/upf.c
```diff
@@ -195,6 +195,24 @@ vnet_upf_nat_pool_add_del (u8 * nwi_name, ip4_address_t start_addr,
...
+
+  policer_main.single_rate = 1;
+  /* Tokens per period, according to police.h should be > 1000 */
+  policer_main.cir_tokens_per_period = 1 << 10;
+  /* scale is set to 10 for init, with given logic that message size is 1 */
+  policer_main.scale = 10;
+  /* we are making top limit of bucket to be high by default */
+  policer_main.current_limit = (1 << 31);
+
+  vec_add1 (vm->upf_pfcp_policer_main, policer_main);
```
ivan4th, line 213: why use a vector here? that's just a single instance. Perhaps could just have a global var (which is already defined below in this commit btw)

upf/upf_cli.c
```diff
@@ -174,6 +174,38 @@ VLIB_CLI_COMMAND (upf_pfcp_show_endpoint_command, static) =
 };
 /* *INDENT-ON* */
 
+static clib_error_t *
+upf_pfcp_policer_set_fn (vlib_main_t * vm,
+			 unformat_input_t * main_input,
+			 vlib_cli_command_t * cmd)
```
ivan4th, line 180: I don't see a definition of CLI command that uses this func. Also, binapi would be nice

ivan4th (changes requested):

upf/upf.c
```diff
@@ -195,6 +195,23 @@ vnet_upf_nat_pool_add_del (u8 * nwi_name, ip4_address_t start_addr,
...
+
+  policer_t *policer_main = &upf_pfcp_policer_main;
+
+  policer_main->single_rate = 1;
+  /* Tokens per period, according to police.h should be > 1000 */
+  policer_main->cir_tokens_per_period = 1 << 10;
+  /* scale is set to 10 for init, with given logic that message size is 1 */
+  policer_main->scale = 10;
+  /* we are making top limit of bucket to be high by default */
+  policer_main->current_limit = (1 << 31);
```
RoadRunnr, line 210: I'm not sure that we should init the policer like this. The normal API seems to use a qos_pol_cfg_params_st struct and pol_logical_2_physical to translate.

upf/upf_pfcp_server.c
```diff
@@ -407,6 +407,13 @@ upf_pfcp_server_send_session_request (upf_session_t * sx,
 				      pfcp_decoded_msg_t * dmsg)
 {
   pfcp_msg_t *msg;
+  upf_main_t *gtm = &upf_main;
+  policer_t *policer_main = gtm->upf_pfcp_policer_main;
+
+  if (vnet_police_packet
+      (policer_main, 1, POLICE_VIOLATE,
```
RoadRunnr, line 414: is think that should `POLICE_CONFORM`

RoadRunnr (changes requested):

upf/upf.c
sergeymatov: this thing has been refactored. Probably another review of the PR is required.

upf/upf.c
sergeymatov: I've refactored PR to use singleton global QoS config that can be configurable via API or CLI
Configuration done via committed PPS and committed burst in ms.
@ivan4th @RoadRunnr please recheck the PR

ivan4th (approved):
It would be nice to have at least some test cases for that, even if just for the binapi itself (although verifying the policier with start-of-traffic reporting trigger is perhaps possible, too).
Other than that, LGTM

RoadRunnr (approved):

PR #236, opened 2022-04-12
Fix 'upf nat pool' CLI command parsing

ivan4th:
Parse min_port/max_port fields
Pass u32* instead of u16* for %u to unformat

### DIFF
--- a/upf/upf.c
+++ b/upf/upf.c
@@ -127,7 +127,10 @@ upf_init_nat_addresses (upf_nat_pool_t * np, ip4_address_t start_addr,
   end = clib_net_to_host_u32 (end_addr.as_u32);
 
   if (start > end)
-    return -1;
+    {
+      clib_warning ("invalid address range for the NAT pool");
+      return -1;
+    }
 
   vec_alloc (np->addresses, end - start + 1);
 
@@ -160,6 +163,15 @@ vnet_upf_nat_pool_add_del (u8 * nwi_name, ip4_address_t start_addr,
       if (p)
 	return VNET_API_ERROR_VALUE_EXIST;
 
+      if (min_port < UPF_NAT_MIN_PORT ||
+	  max_port > UPF_NAT_MAX_PORT || min_port > max_port)
+	{
+	  clib_warning
+	    ("Invalid port range for the NAT pool (must be within %u - %u)",
+	     UPF_NAT_MIN_PORT, UPF_NAT_MAX_PORT);
+	  return VNET_API_ERROR_INVALID_ARGUMENT;
+	}
+
       pool_get (gtm->nat_pools, nat_pool);
 
       if (upf_init_nat_addresses (nat_pool, start_addr, end_addr))
@@ -171,8 +183,8 @@ vnet_upf_nat_pool_add_del (u8 * nwi_name, ip4_address_t start_addr,
       nat_pool->name = vec_dup (name);
       nat_pool->network_instance = vec_dup (nwi_name);
       nat_pool->port_block_size = port_block_size;
-      nat_pool->min_port = UPF_NAT_MIN_PORT;
-      nat_pool->max_port = UPF_NAT_MAX_PORT;
+      nat_pool->min_port = min_port;
+      nat_pool->max_port = max_port;
       nat_pool->max_blocks_per_addr =
 	(u16) ((nat_pool->max_port - nat_pool->min_port) / port_block_size);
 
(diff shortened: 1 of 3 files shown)


upf/upf_cli.c
```diff
@@ -253,15 +253,12 @@ upf_nat_pool_add_del_command_fn (vlib_main_t * vm,
   u8 *nwi_name;
   u8 *nwi_s = 0;
   ip4_address_t start, end;
-  u16 min_port;
-  u16 max_port;
-  u16 port_block_size;
+  u32 min_port = UPF_NAT_MIN_PORT;
```
sergeymatov, line 256: Changing to u32 to fit unformat functions looks good
But let's also change generic usage of min/max ports here.

upf/upf_cli.c
```diff
@@ -253,15 +253,12 @@ upf_nat_pool_add_del_command_fn (vlib_main_t * vm,
...
+  u32 max_port = UPF_NAT_MAX_PORT;
+  u32 port_block_size;
   u8 is_add = 1;
   int rv;
 
-  min_port = UPF_NAT_MIN_PORT;
```
sergeymatov, line 262: don't set it here

sergeymatov (changes requested):

upf/upf_cli.c
```diff
@@ -272,6 +269,10 @@ upf_nat_pool_add_del_command_fn (vlib_main_t * vm,
 	;
       else if (unformat (line_input, "block_size %u", &port_block_size))
 	;
+      else if (unformat (line_input, "min_port %u", &min_port))
```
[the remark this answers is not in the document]
ivan4th: Together with initializing `min_port` and `max_port` to 0, this will cause `nat_pool->max_port` to be `0` if it is not specified, right? That doesn't sound correct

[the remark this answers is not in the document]
sergeymatov: Yep
so, with keeping assignment like you did, but with addition of this code lines?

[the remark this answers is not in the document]
ivan4th: I'd also change underscores to dashes in the command (e.g. `min_port` to `min-port`) for the sake of consistency (we have breaking changes in NAT setup commands in any case, right?)
And add a check for `min_port <= max_port`, do you agree with that?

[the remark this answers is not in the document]
sergeymatov: `min_port <= max_port` should be guaranteed by assignment in upf.c with min/max constants
extra check is redundant
dashes or underscore - let's make a separate PR for this

sergeymatov (approved):

upf/upf.c
```diff
@@ -160,6 +163,15 @@ vnet_upf_nat_pool_add_del (u8 * nwi_name, ip4_address_t start_addr,
       if (p)
 	return VNET_API_ERROR_VALUE_EXIST;
 
+      min_port = clib_max (UPF_NAT_MIN_PORT, min_port);
```
RoadRunnr, line 166: I think you should unconditionally overwrite the min/max port. Instead check and report an error if they exceed the limits.

upf/upf_cli.c
```diff
@@ -253,15 +253,12 @@ upf_nat_pool_add_del_command_fn (vlib_main_t * vm,
   u8 *nwi_name;
   u8 *nwi_s = 0;
   ip4_address_t start, end;
-  u16 min_port;
-  u16 max_port;
-  u16 port_block_size;
+  u32 min_port = UPF_NAT_MIN_PORT;
+  u32 max_port = UPF_NAT_MAX_PORT;
```
RoadRunnr, line 257: `vnet_upf_nat_pool_add_del` takes u16 for min, max and block size. Passing the u32 values into it will truncate the high bits.
The CLI is debug tool, so I don't really mind this. But it might not be what you had in mind. Checking the values for a 16bit overflow might be sensible.

RoadRunnr (changes requested):

upf/upf.c
ivan4th: Fixed

upf/upf_cli.c
ivan4th: Added extra check

upf/upf_cli.c
```diff
@@ -285,15 +286,27 @@ upf_nat_pool_add_del_command_fn (vlib_main_t * vm,
...
 
-  nwi_name = upf_name_to_labels (nwi_s);
-  vec_free (nwi_s);
+  /*
+   * Extra port range check here because port values are parsed into
+   * u32 instead of u16
+   */
+  if (min_port < UPF_NAT_MIN_PORT ||
+      max_port > UPF_NAT_MAX_PORT || min_port > max_port)
+    error = clib_error_return (0, "Invalid port range");
```
RoadRunnr, line 295: I would have use `goto done` here instead of the else block, but that is only a aesthetic thing.

RoadRunnr (approved):

PR #238, opened 2022-04-12
Add UPG_BUILDENV_EXTRA_DIR env var for the dockerized buildenv

ivan4th:
This helps with using custom versions of Go libraries for the tests


korroot (approved):
LGTM

PR #240, opened 2022-04-20
Fix crash upon PSDBU report triggered on a multi-buffer frame

ivan4th:
Make sure that the session is not deleted multiple times

### DIFF
--- a/upf/upf_pfcp_server.c
+++ b/upf/upf_pfcp_server.c
@@ -1210,9 +1210,21 @@ static uword
 		  (upf_event_urr_hdr_t *) vec_header (uev,
 						      sizeof
 						      (upf_event_urr_hdr_t));
-		upf_session_t *sx;
+		upf_session_t *sx = 0;
 
-		sx = pool_elt_at_index (gtm->sessions, ueh->session_idx);
+		/*
+		 * The session could have been freed or replaced after
+		 * this even has been queued, but before it has been
+		 * handled. So we check if the pool entry is free, and
+		 * also verify that CP-SEID in the session matches
+		 * CP-SEID in the event.
+		 */
+		if (!pool_is_free_index (gtm->sessions, ueh->session_idx))
+		  sx = pool_elt_at_index (gtm->sessions, ueh->session_idx);
+
+		if (!sx || sx->cp_seid != ueh->cp_seid)
+		  goto next_ev_urr;
+
 		if (!(ueh->status & URR_DROP_SESSION))
 		  {
 		    upf_debug
@@ -1236,6 +1248,7 @@ static uword
 		    pfcp_free_session (sx);
 		  }
 
+	      next_ev_urr:
 		vec_free_h (uev, sizeof (upf_event_urr_hdr_t));
 	      }
 	    break;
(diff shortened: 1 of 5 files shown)


upf/upf_pfcp_server.h
```diff
@@ -110,6 +110,7 @@ typedef struct
 typedef struct
 {
   uword session_idx;
+  u64 cp_seid;
```
mgumz, line 113: the increase of the data structure by 8 byte (from 21 to 29) … does it change the memory impact of the running instance for 10k 100k 200k sessions in a noticeable way?

ivan4th: This structure is only used when briefly queueing Session Report Requests, so as not to send them right away while handling the user plane packets, and these requests aren't generated often enough to keep more than a few dozen or so in the queue. So I think this will likely have no impact.

upf/upf_pfcp_server.c
```diff
@@ -1212,7 +1212,20 @@ static uword
...
 
+		/*
+		 * The session could have been freed or replaced after
+		 * this even has been queued, but before it has been
+		 * handled. So we check if the pool entry is free, and
+		 * also verify that CP-SEID in the session matches
+		 * CP-SEID in the event.
+		 */
+		if (pool_is_free_index (gtm->sessions, ueh->session_idx))
+		  continue;
```
RoadRunnr, line 1223: The continue skips over the `vec_free_h` at the end. Wouldn't this cause a leak of the ueh entry?

RoadRunnr (commented):
This whole problem makes we wonder if the current URR event queuing model should be changed (but not in this bugfix).
It might be worth having a outstanding URR events queue per session and only queue a trigger to process the URRs to the PFCP server.
That way additional events can be added to the URR queue for a session while the trigger is in the queue. Of course, some mutual exclusion on the per session URR event buffer would be needed.

ivan4th:
I agree concerning the per-session URR event queue (to be done later)

upf/upf_pfcp_server.c
ivan4th: Thanks for noticing this problem. Adjusted the code slightly to avoid the leak

upf/upf_pfcp_server.c
```diff
@@ -1210,9 +1210,24 @@ static uword
...
+		 * CP-SEID in the event.
+		 */
+		if (!pool_is_free_index (gtm->sessions, ueh->session_idx))
+		  sx = pool_elt_at_index (gtm->sessions, ueh->session_idx);
+
+		if (!sx || sx->cp_seid != ueh->cp_seid)
+		  {
+		    vec_free_h (uev, sizeof (upf_event_urr_hdr_t));
+		    continue;
+		  }
```
RoadRunnr, line 1229: Those lines are fine, still I would have coded this as:
```suggestion
		if (!sx || sx->cp_seid != ueh->cp_seid)
		  {
		    goto next_ev_urr;
		  }
```
with a the `goto` label` place right before the free.

That way it would be harder to accidentally use different cleanups for both cases.

RoadRunnr (approved):
LGTM

upf/upf_pfcp_server.c
ivan4th: Updated the code to use `goto`, as indeed we might introduce a hard-to-catch bug later with different cleanup code paths

RoadRunnr (approved):

PR #248, opened 2022-05-18
e2e: fix duplicate VPP processes during IPFIX tests

ivan4th:
This was causing excess memory consumption during E2E test runs,
as well as making PCAP artifacts useless for these test cases
Also, set E2E_PARALLEL_NODES back to 10 for the CI.


korroot (approved):
lgtm

PR #249, opened 2022-05-18
Use fib_table_get_table_id to get table ids from fib indices

ivan4th:
Don't introduce our own func as there's a standard VPP one


sergeymatov (approved):
LGTM

PR #253, opened 2022-05-19
ipfix: use DeltaCount fields instead of initiator/responder fields

ivan4th:
Unlike {initiator,responder}{Packets,Octets} packetDeltaCount /
octetDeltaCount fields depend on the direction, and also the octet
delta counts include L3/L4 headers.

Stacked on top of #252 (merge this one before #252)


sergeymatov (approved):
LGTM

PR #255, opened 2022-05-19
ipfix: reflect forwarding policy id in VRFname

ivan4th:
When a forwarding policy is in use for the flow (upload direction), it
will be reflected in `VRFname` field in IPFIX reports for both
directions of that flow.


sergeymatov (approved):
looks good

PR #256, opened 2022-05-19
ipfix: add ipfix_report_interval field to NWI

ivan4th:
The field specifies minimum per-flow IPFIX reporting interval in
seconds. If the field is set to 0 or ~0, the default value of 5
seconds is used. It's possible to set the value both through the
binapi and CLI.

This PR is stacked on top of #255. This PR must be merged first.


ivan4th:
closing, created by mistake

PR #257, opened 2022-05-19
ipfix: add ipfix_report_interval field to NWI

ivan4th:
The field specifies minimum per-flow IPFIX reporting interval in
seconds. If the field is set to 0 or ~0, the default value of 5
seconds is used. It's possible to set the value both through the
binapi and CLI.
Also, resolve test issues from #255

### DIFF
--- a/upf/upf.api
+++ b/upf/upf.api
@@ -237,6 +237,7 @@ autoreply define upf_nwi_add_del {
   u32 ip6_table_id;
   u8 ipfix_policy[64];
   vl_api_address_t ipfix_collector_ip;
+  u32 ipfix_report_interval;
   u32 observation_domain_id;
   u8 observation_domain_name[256];
   u64 observation_point_id;
--- a/upf/upf.api
+++ b/upf/upf.api
@@ -252,6 +252,10 @@ define upf_nwi_details {
   u32 ip6_table_id;
   u8 ipfix_policy[64];
   vl_api_address_t ipfix_collector_ip;
+  u32 ipfix_report_interval;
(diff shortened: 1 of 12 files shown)


upf/upf.api
```diff
@@ -237,6 +237,7 @@ autoreply define upf_nwi_add_del {
   u32 ip6_table_id;
   u8 ipfix_policy[64];
   vl_api_address_t ipfix_collector_ip;
+  u32 ipfix_report_interval;
```
sergeymatov, line 240: So, this IPFIX-related fields won't be added into nwi_details?

upf/upf_ipfix.c
```diff
@@ -46,11 +46,14 @@
   do { } while (0)
 #endif
 
+/* Default report interval in seconds */
+#define UPF_IPFIX_DEFAULT_REPORT_INTERVAL 5
```
sergeymatov, line 50: why it's traveled from header file to here?

sergeymatov (approved):
LGTM overal

upf/upf.api
ivan4th: Thanks, forgot to add them there. Will post an update shortly

upf/upf_ipfix.c
ivan4th: It's not used outside this C file, and there's no reason to have almost whole UPG recompiled when one changes this constant

upf/upf.api
ivan4th: Added

PR #262, opened 2022-05-23
ipfix: always use origin egress NWi for IPFIX settings

ivan4th:
IPFIX policy can differ for ingress/egress directions of the same
flow (e.g. b/c set in FARs by the control plane), but other IPFIX
settings, including collector IP and observation domain/point id/name,
should always be set for the egress NWI of the flow's origin
direction.


sergeymatov (approved):
LGTM

PR #264, opened 2022-05-26
[VPP-Base] Update VPP-Base to support vpp interface rename

sergeymatov:
New VPP-base version contains a fix for "set interface name" CLI
or binapi call that also renames statistic counters is statseg.


korroot:
Can you rename the title of this PR to describe what new vpp-base brings to the game?
Something like: `Update vpp-base with interface rename fix`

korroot (approved):
LGTM

PR #270, opened 2022-06-02
Fix TCP RST crash in VPP

sergeymatov:
Fix including following changes:
- bump VPP base image to include fix for NAT-ED code checksum assertion for TCP RST packet
- Set up `rx_sw_if_index` in `upf-proxy-input` to allow VPP TCP stack properly set up output interface index for RST packet
- Create E2E test to verify that Volume quota exhaust and adjusted redirect rules for session won't lead UPG to crash


korroot (approved):
LGTM

PR #273, opened 2022-06-07
ipfix: fix context handling

ivan4th:
Don't drop IPFIX contexts when there are no flows, retain them
while the NWI is still present.
Don't crash upon bad IPFIX collector IP spec.


sergeymatov (approved):

korroot (approved):
LGTM

PR #277, opened 2022-06-16
Remove related flows together with PFCP sessions

ivan4th:
This is needed to avoid losing IPFIX reports for the flows that still
exist when a PFCP session is deleted.
As part of the change, flowtable hooks are replaced with flowtable
events which are more flexible (multiple handlers per event).
Possible inconsistencies related to flows that don't correspond to any
active session are now being avoided.

### DIFF
--- a/test/e2e/ipfix_e2e.go
+++ b/test/e2e/ipfix_e2e.go
@@ -41,8 +41,9 @@ func describeIPFIX(title string, mode framework.UPGMode, ipMode framework.UPGIPM
 
 				ginkgo.It("doesn't send IPFIX reports", func() {
 					v.verifyIPFIX(ipfixVerifierCfg{
-						trafficCfg: &traffic.UDPPingConfig{},
-						protocol:   layers.IPProtocolUDP,
+						trafficCfg:  &traffic.UDPPingConfig{},
+						protocol:    layers.IPProtocolUDP,
+						noTemplates: true,
 					})
 					v.verifyNoRecs()
 				})
@@ -158,6 +159,7 @@ func describeIPFIX(title string, mode framework.UPGMode, ipMode framework.UPGIPM
 						farTemplate: "none",
 						trafficCfg:  &traffic.UDPPingConfig{},
 						protocol:    layers.IPProtocolUDP,
+						noTemplates: true,
 					})
 					v.verifyNoRecs()
 				})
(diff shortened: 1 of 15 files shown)


ivan4th:
A bit of explanation:

* there was a “hook” mechanism that enabled to lessen circular
  dependencies between the flowtable code and the rest of UPG; the
  problem with it was that it only allowed one handler per hook, which
  became inconvenient with the new changes.  It has been replaced with
  'events', with some added comments:
  https://github.com/travelping/upg-vpp/blob/56b2959241ebd859d5cf07c77f4c5f8180f2e539/upf/flowtable.h#L237-L267
* Sessions' flows are organized in a singly-linked list. Each flow
  points to the next one belonging to the same session:
  https://github.com/travelping/upg-vpp/blob/56b2959241ebd859d5cf07c77f4c5f8180f2e539/upf/flowtable.h#L168
  and the session contains the index of the first flow:
  https://github.com/travelping/upg-vpp/blob/56b2959241ebd859d5cf07c77f4c5f8180f2e539/upf/upf.h#L800
* when a flow is allocated, it's linked into the chain by making it
  point to the previously-first flow for the session, and then making
  session's first flow index point to it:
  https://github.com/travelping/upg-vpp/blob/56b2959241ebd859d5cf07c77f4c5f8180f2e539/upf/upf_flow_node.c#L225-L237
* when a flow is deleted, first FLOW_EVENT_REMOVE occurs which is
  handled by the IPFIX code to send the final report, if needed
  https://github.com/travelping/upg-vpp/blob/56b2959241ebd859d5cf07c77f4c5f8180f2e539/upf/upf_ipfix.c#L492
  and also by the proxy
  https://github.com/travelping/upg-vpp/blob/56b2959241ebd859d5cf07c77f4c5f8180f2e539/upf/upf_proxy.c#L1403
  so as to remove any connections related to this flow. There's some
  complication with half-open active-side connections, when such one
  is present for the flow, the proxy session is unlinked from the flow
  and kept till the callback is invoked, which will then proceed with
  further cleanup.
* after all FLOW_EVENT_REMOVE handlers are called, FLOW_EVENT_UNLINK
  is invoked which is handled here
  https://github.com/travelping/upg-vpp/blob/56b2959241ebd859d5cf07c77f4c5f8180f2e539/upf/upf_pfcp.c#L1169-L1208
  Basically, it removes the flow from the linked list of the flows
  belonging to the particular session. There's a bit of compromise in
  this part: removing the flow costs `O(n)`, `n` being the number of
  flows per session. We could use a doubly-linked list instead, so as to
  have `O(1)` during flow deletion, but that would further enlarge the
  flowtable entries, decreasing performance.  (Alternative, maybe
  implement later: keep flow list linkage as a separate vector) The
* in `pfcp_free_session`, all of the flows belonging to it are
  forcibly removed:
  https://github.com/travelping/upg-vpp/blob/56b2959241ebd859d5cf07c77f4c5f8180f2e539/upf/upf_pfcp.c#L1140-L1151
  This will invoke `FLOW_EVENT_REMOVE` and `FLOW_EVENT_UNLINK` events
  for each flow, making sure IPFIX reports are sent and proxy
  connections are dealt with.

PR #281, opened 2022-06-17
e2e: interrupt VPP process if it becomes unresponsive

ivan4th:
In case if VPP process hangs, try sending it SIGINT to see
the stack trace in GDB


korroot (approved):
LGTM

PR #294, opened 2022-07-18
Enhancement: Logging for Operations

Haepaxlog:
If this is works, it should give some more information in regards to pfcp_association concerning operations 

logs:
- established Association
- released Association
- suspicious Association (when resending heartbeats after expired heartbeat timers)
- lost Asssociation

### DIFF
--- a/upf/upf_pfcp.c
+++ b/upf/upf_pfcp.c
@@ -53,6 +53,8 @@
   do { } while (0)
 #endif
 
+#define upf_pfcp_associnfo(...) vlib_log_info
+
 upf_main_t upf_main;
 qos_pol_cfg_params_st pfcp_rate_cfg_main;
 
@@ -444,6 +446,10 @@ pfcp_new_association (session_handle_t session_handle,
   vlib_increment_simple_counter (&gtm->upf_simple_counters[UPF_ASSOC_COUNTER],
 				 vlib_get_thread_index (), 0, 1);
 
+  upf_pfcp_associnfo
+    ("PFCP Association Established: Node %U, Local IP %U, Remote IP %U\n",
+     format_node_id, &n->node_id, format_ip46_address, &n->lcl_addr,
+     IP46_TYPE_ANY, format_ip46_address, &n->rmt_addr, IP46_TYPE_ANY);
   return n;
 }
 
@@ -453,6 +459,11 @@ pfcp_release_association (upf_node_assoc_t * n)
   upf_main_t *gtm = &upf_main;
   u32 node_id = n - gtm->nodes;
   u32 idx = n->sessions;
+
+  upf_pfcp_associnfo
+    ("PFCP Association Released: Node %U , Local IP %U, Remote IP %U\n",
+     format_node_id, &n->node_id, format_ip46_address, &n->lcl_addr,
+     IP46_TYPE_ANY, format_ip46_address, &n->rmt_addr, IP46_TYPE_ANY);
 
   switch (n->node_id.type)
     {
@@ -466,8 +477,6 @@ pfcp_release_association (upf_node_assoc_t * n)
       vec_free (n->node_id.fqdn);
       break;
     }
-
-  upf_debug ("pfcp_release_association idx: %u");
 
   while (idx != ~0)
     {
--- a/upf/upf_pfcp_server.c
+++ b/upf/upf_pfcp_server.c
@@ -56,6 +56,8 @@
   do { } while (0)
 #endif
 
+#define upf_pfcp_associnfo(...) vlib_log_info
+
 static void upf_pfcp_make_response (pfcp_msg_t * resp, pfcp_msg_t * req);
 static void restart_response_timer (pfcp_msg_t * msg);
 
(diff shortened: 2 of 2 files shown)


RoadRunnr:
I'm pretty sure this will explode when you beyond e few hundred establishment requests per second. You can to that in a debug image, but I would not want this in a production image.

mgumz:
@RoadRunnr do you expect 100 node-associations per second?

upf/upf_pfcp_server.c
```diff
@@ -394,6 +396,13 @@ request_t1_expired (u32 seq_no)
 	  && !pool_is_free_index (gtm->nodes, msg->node))
 	{
 	  upf_node_assoc_t *n = pool_elt_at_index (gtm->nodes, msg->node);
+
+	  upf_pfcp_associnfo ("PFCP Association Lost: Node {%s}, Node IP {%U}, Local IP %U, Remote IP %U\n", 
+			      n->node_id.fqdn, 
+			      format_ip46_address, &n->node_id.ip, IP46_TYPE_ANY,
```
RoadRunnr, line 402: You can't just assume that IP and FQDN are set in `node_id`. You need to check `node_id.type` and then either print the IP or the FQDN.
Luckily `format_node_id` already does that and can be used here.

upf/upf_pfcp_server.c
```diff
@@ -1139,6 +1148,12 @@ void upf_server_send_heartbeat (u32 node_idx)
   SET_BIT (req->grp.fields, HEARTBEAT_REQUEST_RECOVERY_TIME_STAMP);
   req->recovery_time_stamp = psm->start_time;
 
+  upf_pfcp_associnfo ("PFCP Association Suspicious: Node {%s}, Node IP {%U}, Local IP %U, Remote IP %U\n", 
+		      n->node_id.fqdn,
```
RoadRunnr, line 1152: Here too, use `format_node_id` instead.

upf/upf_pfcp_server.c
```diff
@@ -1225,6 +1240,7 @@ static uword
 
 		    msg = pfcp_msg_pool_add (psm, tx);
 		    upf_pfcp_server_send_request (msg);
+
```
RoadRunnr, line 1243: unrelated white space change, please remove

RoadRunnr (changes requested):
you PR is failing the VPP checkstyle. Please run `hack/checkstyle.sh` before commiting.

upf/upf_pfcp.c
```diff
@@ -467,7 +480,7 @@ pfcp_release_association (upf_node_assoc_t * n)
       break;
     }
 
-  upf_debug ("pfcp_release_association idx: %u");
```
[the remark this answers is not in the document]
ivan4th: There's a bug indeed (although the macro expands to an empty statement by default).
Perhaps we should delete it instead of commenting out

upf/upf_pfcp_server.c
Haepaxlog: Yeah you're right, I reckoned it would be more efficient not doing the check and just logging a default value, but that probably wouldn't have worked. Thanks for the heads up on format_node_id. Consider it fixed.

upf/upf_pfcp_server.c
```diff
@@ -1309,7 +1322,7 @@ static uword
 
 	  switch (psm->expired[i] >> 24)
 	    {
-	    case 0 ... 0x7f:
+	    case 0...0 x7f:
```
RoadRunnr, line 1325: this change shouldn't be in here

upf/upf_pfcp_server.c
```diff
@@ -1126,9 +1134,9 @@ void upf_pfcp_server_deferred_free_msgs_by_node (u32 node)
 void upf_server_send_heartbeat (u32 node_idx)
 {
   pfcp_server_main_t *psm = &pfcp_server_main;
-  pfcp_decoded_msg_t dmsg = {
-    .type = PFCP_HEARTBEAT_REQUEST
-  };
+  pfcp_decoded_msg_t dmsg =
+  {
+  .type = PFCP_HEARTBEAT_REQUEST};
```
RoadRunnr, line 1139: this change shouldn't be in here

upf/upf_pfcp.c
```diff
@@ -2590,7 +2599,7 @@ process_qers (vlib_main_t * vm, upf_session_t * sess,
   {
     upf_qer_t *qer = pfcp_get_qer_by_id (r, *qer_id);
     upf_qer_policer_t *pol;
-    u32 col __attribute__((unused));
+    u32 col __attribute__ ((unused));
```
RoadRunnr, line 2602: this change shouldn't be in here

upf/upf_pfcp.c
```diff
@@ -2565,7 +2574,7 @@ process_urrs (vlib_main_t * vm, upf_session_t * sess,
 
 u32
 process_qers (vlib_main_t * vm, upf_session_t * sess,
-	      struct rules *r,
+	      struct rules * r,
```
RoadRunnr, line 2577: this change shouldn't be in here

RoadRunnr (changes requested):
It looks like you reformatted the code with an indent version that produced unexpected results.
Only the one used by CI is the correct one.

In any case, do not blindly commit unrelated formatting changes.

upf/upf_pfcp_server.c
```diff
@@ -1139,6 +1147,11 @@ void upf_server_send_heartbeat (u32 node_idx)
   SET_BIT (req->grp.fields, HEARTBEAT_REQUEST_RECOVERY_TIME_STAMP);
   req->recovery_time_stamp = psm->start_time;
 
+  upf_pfcp_associnfo
```
sergeymatov, line 1150: Aint this going to happen regardlessly if association is suspicious or not? It seems it will be happen for every normal heartbeat.

sergeymatov (changes requested):

upf/upf_pfcp_server.c
Haepaxlog: Yeah, you're right. It will only be suspicious when sent by a heartbeat timeout. So maybe it would be better to log it in that case but not in send_heartbeat?

PR #296, opened 2022-07-21
Fix and improve 'show upf session'

ivan4th:
Fix filtering by hex UP SEIDs (`show upf session up seid 0x...`)
Add `show upf session 0x... flows` to show the flows related to the
session.
Limit `show upf session` to displaying 100 sessions by default.
This can be overridden via `show upf session limit N`
(`limit 0` stands for unlimited)

### DIFF
--- a/upf/upf_cli.c
+++ b/upf/upf_cli.c
@@ -37,6 +37,9 @@
 #include <vnet/fib/fib_path_list.h>
 #include <vnet/fib/fib_walk.h>
 
+#define DEFAULT_MAX_SHOW_UPF_SESSIONS 100
+#define HARD_MAX_SHOW_UPF_SESSIONS    10000
+
 #if CLIB_DEBUG > 1
 #define upf_debug clib_warning
 #else
@@ -1082,17 +1085,28 @@ VLIB_CLI_COMMAND (upf_show_gtpu_endpoint_command, static) =
 };
 /* *INDENT-ON* */
 
+typedef struct
+{
+  vlib_main_t *vm;
+  u64 seid;
+  bool filtered;
+  u32 limit;
+} flows_out_arg_t;
+
 static int
 upf_flows_out_cb (clib_bihash_kv_48_8_t * kvp, void *arg)
 {
   flowtable_main_t *fm = &flowtable_main;
-  vlib_main_t *vm = (vlib_main_t *) arg;
+  flows_out_arg_t *arg_value = (flows_out_arg_t *) arg;
+  flow_key_t *key = (flow_key_t *) & kvp->key;
   flow_entry_t *flow;
 
   flow = pool_elt_at_index (fm->flows, kvp->value);
-  vlib_cli_output (vm, "%U", format_flow, flow);
+  if (!arg_value->filtered || arg_value->seid == key->seid)
+    vlib_cli_output (arg_value->vm, "%U", format_flow, flow);
 
-  return (BIHASH_WALK_CONTINUE);
+  return arg_value->limit != ~0 && !--arg_value->limit ?
+    BIHASH_WALK_STOP : BIHASH_WALK_CONTINUE;
 }
 
 static clib_error_t *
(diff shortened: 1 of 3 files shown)


hwinkel:
Do we have by IMSI as well, as we signal the imsi now down to UPG  from SMC As part of PFCP IEs

RoadRunnr:
> Do we have by IMSI as well, as we signal the imsi now down to UPG from SMC As part of PFCP IEs

This would require additional hashes to index the session by IMSI/IMEI/MSISDN/what-not. The overhead and potential for problems is IMHO not worth it.

upf/upf_cli.c
```diff
@@ -1197,7 +1221,7 @@ VLIB_CLI_COMMAND (upf_show_session_command, static) =
 {
   .path = "show upf session",
   .short_help =
-  "show upf session",
+  "show upf session [up seid 0x...] [0x... flows] [limit N]",
```
RoadRunnr, line 1224: do not mix prefix and postfix parameters. VPP cli command always use the prefix form, stick with that please.

The VPP CLI is not operation tool! The usability of the command is not really important. But, you create a potential of messing up the command parsing if you mix postfix and prefix arguments like that.

RoadRunnr: Also, limiting the output is pretty useless. Either you know that it is safe to list all or there are too many session that listing even a part of them is useless anyhow.

I would suggest to change the command to either display a single session or all. Use a special flag to force all, e.g.:
```
"show upf session [ up seid N [flows N] | all]"
```

ivan4th:
Well, risk-wise a dumber approach of traversing the whole session pool looking for a particular IMSI might not be that bad, if it is not done repeatedly in a tight loop, which typically doesn't happen during CLI debug sessions.

From the experience, `show upf flows` doesn't cause disruptions most of the time despite going over the whole flowtable, and the session pool usually contains fewer entries. The risk of the disruption from `show upf session` stems from generating the output, not from examining the data structures.

sergeymatov (approved):
lgtm

upf/upf_cli.c
korroot: If we have like 1k sessions and want to limit the output, so running like `show upf session limit 100` will always show us first 100 sessions, and the rest 900 will not be accessible - either there should be paging (like `limit 100-200`) or just remove the limit... 

But maybe if first entries can say us something meaningful than it can stay, what we can gain by seeing first sessions?

korroot: When I reread the commit description the help should be:
```
show upf session [up seid 0x...][up seid 0x... flows][limit N]
```
Right?

ivan4th:
Updated the PR. The syntax now is
```
show upf session [up seid 0x... [flows]] [limit N]
```

Indeed, I think showing just 100 sessions are still useful b/c of 2 possibilities
1) pure debug/test settings where the actual number of sessions is less than 100
2) to get some clues from a number of sessions in the system

And having the limit in place by default is also critical b/c `show upf session` is known to have caused actual UPG crashes in prod due to the amount of output.

I've also added `limit N` option to `show upf flows`. There's no limit by default b/c as far as I know, UPG never ever has crashed in prod when executing this command. But the limit may be useful for synthetic situations where unusually high number of flows is created using TPLoad or TRex, for example.

upf/upf_cli.c
```diff
@@ -1108,34 +1121,33 @@ upf_show_session_command_fn (vlib_main_t * vm,
...
+	  else if (unformat (line_input, "up seid 0x%llx", &up_seid))
 	    has_up_seid = 1;
-	  else if (unformat (line_input, "up seid 0x%lx", &up_seid))
+	  else if (unformat (line_input, "up seid %lu", &up_seid))
 	    has_up_seid = 1;
 	  else if (unformat (line_input, "debug"))
 	    debug = 1;
-#if FLOWTABLE_TODO
-	  else if (unformat (line_input, "%lu flows", &up_seid))
+	  else if (unformat (line_input, "0x%llx flows", &up_seid))
```
RoadRunnr, line 1145: what is this used for? all cases for parsing up_seid should already be covered.

upf/upf_cli.c
```diff
@@ -1147,28 +1159,45 @@ upf_show_session_command_fn (vlib_main_t * vm,
...
+    }
+
   if (has_flows)
     {
+      u32 cpu_index;
+      flowtable_main_t *fm = &flowtable_main;
+      vlib_thread_main_t *tm = vlib_get_thread_main ();
+      flows_out_arg_t arg = {.vm = vm,.seid = up_seid,.filtered =
+	  true,.limit = limit
+      };
```
RoadRunnr, line 1176: Strange that this passes through the format checks, I would expected that there should be a space after a comma.

upf/upf_cli.c
```diff
@@ -1108,34 +1121,33 @@ upf_show_session_command_fn (vlib_main_t * vm,
...
-#if FLOWTABLE_TODO
-	  else if (unformat (line_input, "%lu flows", &up_seid))
+	  else if (unformat (line_input, "0x%llx flows", &up_seid))
 	    has_flows = 1;
-	  else if (unformat (line_input, "0x%lx flows", &up_seid))
+	  else if (unformat (line_input, "flows", &up_seid))
 	    has_flows = 1;
-#endif
+	  else if (unformat (line_input, "limit %u", &limit))
+	    ;
```
RoadRunnr, line 1150: should we enforce a hard max for limit? Currently setting limit to 0 would effectively mean `unlimited`

RoadRunnr (changes requested):

upf/upf_cli.c
ivan4th: Hmm, it looks like it was actually `indent` which formatted it this way, and I overlooked it. Split it into separate lines

upf/upf_cli.c
ivan4th: Added the hard limit of 10K sessions

RoadRunnr (approved):

PR #297, opened 2022-08-03
Add association logging and fix use-after-free

ivan4th:
This adds some fixes to the changes from #294.
The code is buildable now. Also, the following is needed in `startup.conf` to enable UPG-related logging
(the class is called `upf` for consistency with other UPG aspects):

```
logging {
  class upf { level info syslog-level info }
}
```

As an alternative, logging can be enabled via CLI
```
set logging class upf level info syslog-level info
```

Without additional syslog config, `syslog-level` setting is needed to see the messages printed to the stderr. Otherwise, they will only be visible in `show logging` command output.

Also, a use-after-free issue is fixed in the PFCP server code, and `pfcp_msg_t` data poisoning is added.

### DIFF
--- a/upf/upf_pfcp_server.c
+++ b/upf/upf_pfcp_server.c
@@ -396,6 +396,7 @@ request_t1_expired (u32 seq_no)
   else
     {
       u8 type = pfcp_msg_type (msg->data);
+      u32 node = msg->node;
 
       upf_debug ("abort...\n");
       // TODO: handle communication breakdown....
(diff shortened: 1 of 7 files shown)


upf/upf_pfcp_server.c
```diff
@@ -414,8 +411,14 @@ request_t1_expired (u32 seq_no)
 	     format_node_id, &n->node_id, format_ip46_address, &n->lcl_addr,
 	     IP46_TYPE_ANY, format_ip46_address, &n->rmt_addr, IP46_TYPE_ANY);
 
+	  /* This will free the message at the end of pfcp_process() loop */
 	  pfcp_release_association (n);
 	}
+      else
+	{
+	  hash_unset (psm->request_q, msg->seq_no);
+	  pfcp_msg_pool_put (psm, msg);
```
RoadRunnr, line 420: Instead of adding else clause and moving the msg release there, wouldn't it be simpler to leave it where it is and to copy the `msg->node` value to local variable and use that in the association setup?

RoadRunnr (approved):

korroot (approved):

PR #303, opened 2022-09-26
[e2e] check if UE traffic with non-assigned IP address is not passing.

sergeymatov:


RoadRunnr (changes requested):
The description of the changeset is invalid.

A UPF should do that the PDRs and FARs tell it to do. If the PDRs permit IPs other then the UE IP, then it is the job of the UPF to forward them. It has not bussines to infer rules that are not setup by the control path.

There are valid use cases, where devices behind a UE send data with source IPs that do not match the UEs IP.

That does not mean the the test is invalid. Ensuring that only source IPs matched by PDRs/FARs are forwarded is a valid check.

However, it would be possible to create PDRs/FARs that permit additional IP ranges as source IPs. In that case, those IPs should be forwarded.

PR #304, opened 2022-09-28
Fix setting of buffer offload flags

mitmitmitm:
Rather than in b->flags, the offload flags are really stored in `vnet_buffer (b)->oflags' and `vnet_buffer_offload_flags_clear' should be used to set them.


mogaika (approved):
LGTM. Good finding. Sorry for long wait...

PR #307, opened 2022-11-14
[CENNSO-72][NAT] Update default MIN and MAX port range for NAT

sergeymatov:
Per-pool configuration option is available via CLI

### DIFF
--- a/test/e2e/upg_e2e.go
+++ b/test/e2e/upg_e2e.go
@@ -2127,7 +2127,7 @@ func verifyPSDBU(m message.Message, numUsageReports int) {
 func setupNAT(f *framework.Framework) {
 	f.VPP.Ctl("nat44 enable sessions 1000")
 	f.VPP.Ctl("set interface nat44 out host-sgi0 output-feature")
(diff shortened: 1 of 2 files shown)


test/e2e/upg_e2e.go
```diff
@@ -2127,7 +2127,7 @@ func verifyPSDBU(m message.Message, numUsageReports int) {
 func setupNAT(f *framework.Framework) {
 	f.VPP.Ctl("nat44 enable sessions 1000")
 	f.VPP.Ctl("set interface nat44 out host-sgi0 output-feature")
-	f.VPP.Ctl("upf nat pool 144.0.0.20 - 144.0.0.120 block_size 512 nwi sgi name testing")
+	f.VPP.Ctl("upf nat pool 144.0.0.20 - 144.0.0.120 block_size 512 nwi sgi name testing min_port 10128")
```
korroot, line 2130: maybe we  can add max_port as well. Is it verified somehow?

sergeymatov: we are not verifying upper border, by adding min_port here I've restored original test check for `10128` port

korroot (approved):
LGTM

PR #316, opened 2023-01-23
remove support for pre 3GPP Rel.16 PFCP and UPF features

RoadRunnr:
The PFCP User Plane IP Resource Information IE has been removed in 3GPP Release 16 and UP based TEID assignment is mandatory since that release as well.

User Plane IP Resource Information where never really useful, therefore dropping it helps us simplifying the code without losing anything.

UP based TEID assignment (FTUP feature flag) is mandatory if multiple SMFs want to control the same UPF. This is also an important feature (which part of why it become mandatory in 3GPP Rel. 16).

### DIFF
--- a/upf/test/test_upf.py
+++ b/upf/test/test_upf.py
@@ -118,7 +118,6 @@ class IPv4Mixin(object):
             "upf nwi name cp vrf 0",
             "upf nwi name epc vrf 100",
             "upf nwi name sgi vrf 200",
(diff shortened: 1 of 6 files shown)


korroot:
As these are not backward-compatible changes, we can merge it for upg-vpp 2.0 release.

sergeymatov (commented):
lgtm but lets hold on this a bit

PR #318, opened 2023-01-25
Fix crash in session delete (sparse_vec_free)

miguelbf-alb:
I am facing several issues when the PFCP session is removed for a given UE, usually vpp/upg is left on a deadlock state and can't be used anymore. Debugging this further the culprit is in the call to `sparse_vec_free`. I guess the vec pointer has to be deferenced. Similar code flow exists on other parts of vpp (e.g. https://github.com/FDio/vpp/blob/9bc72ac8de6dcc74aa82fa6c3223e9f93b2dc3c2/src/vnet/udp/udp.c#L491-L510)
Runtime testing this change and the deadlock is gone

regards

### DIFF
--- a/upf/upf_pfcp.c
+++ b/upf/upf_pfcp.c
@@ -1162,7 +1162,7 @@ pfcp_free_session (upf_session_t * sx)
   for (size_t i = 0; i < ARRAY_LEN (sx->rules); i++)
     pfcp_free_rules (sx, i);
 
-  sparse_vec_free (sx->teid_by_chid);
+  sparse_vec_free (*sx->teid_by_chid);
 
   if (sx->nat_addr)
     upf_delete_nat_binding (sx);


sergeymatov:
From the code example you've provided `port_nh_sv` is still a pointer to vector, can be checked in lines 486 or 501 of file given.
This code line will executed every time when PFCP session is about to be deleted, so it's covered by multiple test cases within the `test/e2e` set.
Dereference is surely not needed, probably you hit different issue.

sergeymatov (changes requested):
sparse_vec_free required pointer to vector being passed to argument.
This won't work, function is safe enough to not to crash on this changes.

miguelbf-alb:
@sergeymatov you're right, the argument on that snippet is actually a pointer to the vec pointer (hence the need for dereferencing it).
Digging a bit more on this the actual root cause of this is on the implementation of  `sparse_vec_free` on vpp 22.02. Updating the function implementation (specially by considering your commit https://github.com/FDio/vpp/commit/ac199fcd9ba16a9dc3657f8ee02c2a2c82a65417) fixes the issue.
Sorry for wasting your time with this one.

PR #319, opened 2023-01-27
[CENNSO-102] Fix UE IP Address Pool Identity IE handling code

ivan4th:
This fixes crashes due to heap corruption in UE IP Address Pool Identity IE encoding and decoding functions.
Also, integration test script is fixed so just one of the test cases can be run instead of the whole test_upf.


sergeymatov (approved):
lgtm

PR #323, opened 2023-03-27
[CENNSO-617][wip] Don't allow UE-UE access

sergeymatov:


RoadRunnr:
I now this is a WIP, but before it goes to far in the wrong direction, I have to point out that UE-UE blocking should be a per-NWI feature and not a UPF global thing.

There is no guarantee that all NWIs/DNNs/APNs on one UPF instance want the same policy. Anything that provides public internet access is likely to want UE-UE blocking, but private APN application might not. Making it a global feature would prohibit running multiple DNNs/APNs on the same UPF.

Also, I must point out that using proper PCC rules (resulting in proper PDR and FAR rules) should already be able to block UE-UE communication.

PR #326, opened 2023-04-24
[CENNSO-688] Fix excess heartbeat requests after reassociation

ivan4th:
When a PFCP association is released and later reestablished, or
established towards another SMF/PGW-C, the old heartbeat timer was not
being stopped, causing excess heartbeat requests to be sent by the UPG.


korroot (approved):
LGTM

PR #329, opened 2023-05-18
feat: binapi impl for set/get node-id

demo-exe:


ivan4th (changes requested):

test/e2e/upg_e2e.go
```diff
@@ -604,6 +604,98 @@ var _ = ginkgo.Describe("CLI debug commands", func() {
 
 // TODO: validate both binapi and CLI against each other
 var _ = ginkgo.Describe("UPG Binary API", func() {
+	ginkgo.Context("for node-id", func() {
+		f := framework.NewDefaultFramework(framework.UPGModeTDF, framework.UPGIPModeV4)
+
+		callSetNodeID := func(req *upf.UpfSetNodeID) (*upf.UpfSetNodeIDReply, error) {
+			reply := &upf.UpfSetNodeIDReply{}
+			err := f.VPP.ApiChannel.SendRequest(req).ReceiveReply(reply)
+			return reply, err
```
ivan4th, line 613: Probably no need to return the `reply` here, it doesn't carry any useful info. You can just use
```suggestion
			return f.VPP.ApiChannel.SendRequest(req).ReceiveReply(reply)
```

test/e2e/upg_e2e.go
```diff
@@ -604,6 +604,98 @@ var _ = ginkgo.Describe("CLI debug commands", func() {
...
+			fqdn := util.EncodeFQDN(fqdnStr)
+			fqdnLen := uint8(len(fqdn))
+
+			// pass 1: IPv4
+			setReq := &upf.UpfSetNodeID{
+				Type: uint8(upf.UPF_NODE_TYPE_IPv4),
+				IP:   ipv4,
+			}
+			_, setErr := callSetNodeID(setReq)
+			gomega.Expect(setErr).To(gomega.BeNil(), "upf_set_node_id")
```
ivan4th, line 637: To be consistent with the most of Go code, you need to just call it `err`. Later you can reuse it still.
If you're suing `:=`, just one of the left-side identifiers needs to be "new". So you can write
```go
			_, err := callSetNodeID(setReq)
```
and then still use 
```go
			out, err := f.VPP.Ctl("show upf node-id")
```

If `out` is previously defined, though, you'll have to do
```go
			out, err := f.VPP.Ctl("show upf node-id")
```
IoW: `:=` needs to be defining "some" new vars. But not all vars need to be new.

Also, in Gomega, for `err` checking you write
```go
			gomega.Expect(setErr).NotTo(gomega.HaveOccurred(), "upf_set_node_id")
```

But with the change suggested above, you can actually just use 
```suggestion
			gomega.Expect(callSetNodeID(setReq)).To(gomega.Succeed(), "upf_set_node_id")
```

test/e2e/upg_e2e.go
```diff
@@ -604,6 +604,98 @@ var _ = ginkgo.Describe("CLI debug commands", func() {
...
+
+			out, err := f.VPP.Ctl("show upf node-id")
+			gomega.Expect(err).NotTo(gomega.HaveOccurred(), "show upf node-id")
+			gomega.Expect(out).To(gomega.ContainSubstring(ipv4.ToIP().String()), "expected node-id")
+
+			getReply, getErr := callGetNodeID()
```
ivan4th, line 643: Also `err` instead of `getErr`

test/e2e/upg_e2e.go
```diff
@@ -604,6 +604,98 @@ var _ = ginkgo.Describe("CLI debug commands", func() {
...
+				IP:   ipv6,
+			}
+			_, setErr = callSetNodeID(setReq)
+			gomega.Expect(setErr).To(gomega.BeNil(), "upf_set_node_id")
+
+			out, err = f.VPP.Ctl("show upf node-id")
+			gomega.Expect(err).NotTo(gomega.HaveOccurred(), "show upf node-id")
+			gomega.Expect(out).To(gomega.ContainSubstring(ipv6.ToIP().String()), "expected node-id")
+
+			getReply, getErr = callGetNodeID()
```
ivan4th, line 665: `err` instead of `getErr`

upf/upf_api.c
```diff
@@ -892,11 +891,94 @@ vl_api_upf_pfcp_heartbeats_get_t_handler (vl_api_upf_pfcp_heartbeats_get_t *
...
+  upf_main_t *sm = &upf_main;
+  vl_api_upf_set_node_id_reply_t *rmp = NULL;
+
+  // if old was fqdn then free it
+
+  switch (mp->type)
+    {
+    case NID_IPv4:
+    case NID_IPv6:
+      sm->node_id.type = mp->type;
```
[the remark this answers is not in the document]
ivan4th: UPD: there's no bug in `vnet_upf_node_id_set()`, looked at it in a wrong way :)

ivan4th (changes requested):
Some minor quirks in the Go test remaining

PR #330, opened 2023-05-23
feat: initial vscode configuration

demo-exe:


.vscode/c_cpp_properties.json
```diff
@@ -1,18 +1,18 @@
...
-            "configurationProvider": "ms-vscode.makefile-tools"
-        }
-    ],
-    "version": 4
+  // for use with remote VSCode inside Docker container
+  "configurations": [
+    {
+      "name": "Linux",
+      "includePath": [
+        "${workspaceFolder}/**"
```
ivan4th, line 7: Q: I'm not an expert in VSCode configs, but shouldn't that include vpp sources?
They reside under `/vpp-src` in the build container

demo-exe: You are right, the fact that includes are copied to /usr/include and hence found by the compiler misled me

hack/buildenv.sh
```diff
@@ -51,10 +52,23 @@ function docker_buildenv {
...
+    docker run -d --rm --name vpp-build-${BUILD_TYPE}-bg --shm-size 1024m \
+          ${priv} \
+          -v $PWD:/src:delegated -v $PWD/vpp-out:/vpp-out \
+          "${opts[@]}" -w /src "${DEV_IMAGE}"
+  else
+    docker run --rm --name vpp-build-${BUILD_TYPE} --shm-size 1024m \
+          ${priv} \
+          -v $PWD:/src:delegated -v $PWD/vpp-out:/vpp-out \
+          "${opts[@]}" -w /src "${DEV_IMAGE}" "$@"
+  fi
```
ivan4th, line 71: I'd rewrite this as follows. I don't think that you need to wait after `docker rm -f`, which kills a running container and then removes it. This is worth doing before any such `docker run`, not only the background one, otherwise it will fail if the container is there.
```suggestion
  name="vpp-build-${BUILD_TYPE}"
  if [[ ${DEVENV_BG} ]]; then
    name="vpp-build-${BUILD_TYPE}-bg"
    opts+=(-d)
  fi
  docker rm -f "${name}"
  docker run --rm --name "${name}" --shm-size 1024m \
         ${priv} \
         -v $PWD:/src:delegated -v $PWD/vpp-out:/vpp-out \
         "${opts[@]}" -w /src "${DEV_IMAGE}" "$@"
```

hack/attach-gdb.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# this script tries to attach to a gdbserver
+
+# wait for gdbserver to start
+while ! pidof gdbserver >& /dev/null; do sleep 1; done
+nsenter <&0 -t $(pgrep gdbserver) -n gdb $@
```
ivan4th, line 6: I'd say using `pidof` instead of `pgrep` could be slightly more correct there. Also, please use `"$@"` instead of just `$@` (without quotes). In this particular case, that hardly matters much, but in other cases, `$@` will not handle `"argument values with spaces"` correctly

hack/buildenv.sh
demo-exe: damn, I did overcomplicate it haha

ivan4th (approved):
LGTM

PR #337, opened 2023-06-13
[CENNSO-919] feat: binapi for ueip pool

demo-exe:


test/e2e/upg_e2e.go
```diff
@@ -604,6 +604,34 @@ var _ = ginkgo.Describe("CLI debug commands", func() {
 
 // TODO: validate both binapi and CLI against each other
 var _ = ginkgo.Describe("UPG Binary API", func() {
+	ginkgo.Context("for upf ueip pool", func() {
+		f := framework.NewDefaultFramework(framework.UPGModeTDF, framework.UPGIPModeV4)
+
+		addPool := func(isAdd bool, nwi_s string, name string) {
```
ivan4th, line 610: Please don't use underscores in Go variable names. `nwi` should do here

upf/upf.api
```diff
@@ -409,3 +409,14 @@ autoreply define upf_tdf_ul_table_add {
...
+
+autoreply define upf_ueip_pool_nwi_add {
+  u32 client_index;
+  u32 context;
+
+  bool is_add;
+  /* 2 null terminated strings concatenated together
+   * 1st - nwi-name, 2nd - identity */
+  u8 names_len;
+  u8 names[names_len];
```
ivan4th, line 421: I think using a single field for both is rather inconvenient. Perhaps we could have
```suggestion
  u8 identity[64];
  u8 nwi_len;
  u8 nwi[nwi_len];
```

I think 64 bytes should be enough for the identity field (zero-terminated)

upf/upf_api.c
```diff
@@ -1061,6 +1061,31 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+  u8 *name = mp->names;
+  u8 *nwi_name;
+
+  // calculate start of second cstring
+  while (*name++);
+
+  vec_validate (nwi_s_vec, name - nwi_s - 1);
+  memcpy (nwi_s_vec, nwi_s, name - nwi_s - 1);
+
+  nwi_name = upf_name_to_labels (nwi_s_vec);
```
ivan4th, line 1083: In other API calls, we don't use `upf_name_to_labels`, expecting the transformation to be done on the client side, so we should be consistent here

upf/upf.api
```diff
@@ -409,3 +409,14 @@ autoreply define upf_tdf_ul_table_add {
...
+};
```
ivan4th, line 422: We also need a binapi call for retrieving the UE IP pools

ivan4th (changes requested):

upf/upf_api.c
```diff
@@ -1061,19 +1061,84 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+  uword *p;
+  upf_ue_ip_pool_info_t *pool = NULL;
+  u8 *identity;
+
+  reg = vl_api_client_index_to_registration (mp->client_index);
+  if (!reg)
+    return;
+
+  /* *INDENT-OFF* */
+  hash_foreach (identity, p, sm->ue_ip_pool_index_by_identity,
```
ivan4th, line 1105: There is `pool_foreach` which iterates over non-free pool entries

upf/upf_api.c
```diff
@@ -1061,19 +1061,84 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+    {
+      u8 nwi_name_len;
+      u8 identity_len;
+
+      p = hash_get_mem (sm->ue_ip_pool_index_by_identity, identity);
+      pool = pool_elt_at_index (sm->ueip_pools, p[0]);
+
+      nwi_name_len = strlen (pool->nwi_name);
```
ivan4th, line 1113: Please check that `nwi_name_len` doesn't exceed the max possible length (64).
Best of all, it should be ensured that NWIs which such long names can't be created (incl. via CLI), and an assertion added here

test/e2e/upg_e2e.go
```diff
@@ -604,6 +604,68 @@ var _ = ginkgo.Describe("CLI debug commands", func() {
...
+			return ret
+		}
+
+		ginkgo.It("adds and removes a pool", func() {
+			addPool(true, "sgi", "mypool")
+			addPool(true, "test", "mypool")
+
+			pools := dumpPools()
+			gomega.Expect(pools).To(gomega.ContainElements(poolPairing{"sgi", "mypool"}, poolPairing{"test", "mypool"}))
+			gomega.Expect(len(pools)).To(gomega.Equal(2))
```
ivan4th, line 660: If you use the suggestion below about `pool_foreach`, the elements will be in the order in which they were added, and you can just check equality of the slices

ivan4th (changes requested):

test/e2e/upg_e2e.go
demo-exe: I don't think the test should break if this order changes, so I'm for leaving it as is unless you insist to change it

ivan4th: Ok, in this case you might rewrite it as 
```suggestion
			gomega.Expect(dumpPools()).To(gomega.ConsitOf(poolPairing{"sgi", "mypool"}, poolPairing{"test", "mypool"}))
```

See Gomega matcher reference here:
https://onsi.github.io/gomega/#provided-matchers

upf/upf_cli.c
```diff
@@ -254,6 +254,14 @@ upf_ueip_pool_add_del_command_fn (vlib_main_t * vm,
...
   vec_free (nwi_s);
 
+  if (strlen (nwi_name) > 64)
+    return clib_error_return (0,
+			      "NWI name(encoded) has to be shorter than 64 bytes");
+
+  if (strlen (name) > 64)
+    return clib_error_return (0,
+			      "UE IP pool name has to be shorter than 64 bytes");
+
```
ivan4th, line 264: `upf_name_to_labels` returns a VPP vector which is not normally null-terminated
You need to use `vec_len` with vectors:
```suggestion
  if (vec_len (nwi_name) > 64)
    return clib_error_return (0,
			      "NWI name(encoded) has to be shorter than 64 bytes");

  if (vec_len (name) > 64)
    return clib_error_return (0,
			      "UE IP pool name has to be shorter than 64 bytes");

```

upf/upf_api.c
```diff
@@ -1061,6 +1061,68 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+  if (!reg)
+    return;
+
+  pool_foreach (pool, sm->ueip_pools)
+  {
+    u8 nwi_name_len;
+    u8 identity_len;
+
+    nwi_name_len = strlen (pool->nwi_name);
+    ASSERT (nwi_name_len < 64);
```
ivan4th, line 1109: 64 is acceptable size and can be produced by the CLI command below

upf/upf_api.c
```diff
@@ -1061,6 +1061,68 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+    identity_len = strlen (pool->identity);
+    ASSERT (identity_len < 64);
```
ivan4th, line 1111: 64 is acceptable size and can be produced by the CLI command below

upf/upf_api.c
```diff
@@ -1061,6 +1061,68 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+  int rv = 0;
+  upf_main_t *sm = &upf_main;
+  vl_api_upf_ueip_pool_nwi_add_reply_t *rmp = NULL;
+
+  u8 *nwi_name = mp->nwi_name;
+  u8 *nwi_name_vec = 0;
+  u8 *identity = mp->identity;
+  u8 *identity_vec = 0;
+  u8 identity_len = strlen (identity);
+  ASSERT (identity_len < 64);
```
ivan4th, line 1077: 64 is acceptable size and can be produced by the CLI command below

ivan4th (changes requested):

upf/upf_api.c
demo-exe: strlen does not include terminator, fixed below lines

upf/upf_api.c
```diff
@@ -1061,6 +1061,68 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+{
+  int rv = 0;
+  upf_main_t *sm = &upf_main;
+  vl_api_upf_ueip_pool_nwi_add_reply_t *rmp = NULL;
+
+  u8 *nwi_name = mp->nwi_name;
+  u8 *nwi_name_vec = 0;
+  u8 *identity = mp->identity;
+  u8 *identity_vec = 0;
+  u8 identity_len = strlen (identity);
```
ivan4th, line 1076: Using `strlen` on a field that is potentially non-null-terminated is not correct

ivan4th (changes requested):

test/e2e/upg_e2e.go
```diff
@@ -604,6 +604,69 @@ var _ = ginkgo.Describe("CLI debug commands", func() {
...
+					break
+				}
+
+				identity := strings.TrimRight(string(msg.Identity), "\x00")
+				nwiName := strings.TrimRight(string(msg.NwiName), "\x00")
+
+				ret = append(ret, poolPairing{
+					nwi:  util.DecodeFQDN([]byte(identity)),
+					name: util.DecodeFQDN([]byte(nwiName)),
+				})
```
RoadRunnr, line 648: there is some confusion going on here:

* the identity is the PFCP "UE IP address Pool Identity", is a) not a `nwi` and b) a plain binary string without any special encoding
* nwiName is the PFCP "Network Instance", encoding rules are somewhat muddy, but it can be either a plain binary string or use DNS label encoding. `{En,De}codeFQDN` appears to only support the DNS label encoding (and not particularly good, but might be good enough for testing)

test/e2e/upg_e2e.go
```diff
@@ -604,6 +604,69 @@ var _ = ginkgo.Describe("CLI debug commands", func() {
...
+
+		addPool := func(isAdd bool, nwi string, name string) {
+			nwiConverted := append(util.EncodeFQDN(nwi), 0)
+			nameConverted := append(util.EncodeFQDN(name), 0)
+			req := &upf.UpfUeipPoolNwiAdd{
+				IsAdd:      isAdd,
+				Identity:   nwiConverted,
+				NwiName:    nameConverted,
+				NwiNameLen: uint8(len(nameConverted) - 1),
+			}
```
RoadRunnr, line 618: same comment as on line 642: https://github.com/travelping/upg-vpp/pull/337/files#r1252815086

upf/upf_api.c
```diff
@@ -1061,6 +1061,78 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+  u8 *nwi_name_vec = 0;
+  u8 *identity = mp->identity;
+  u8 *identity_vec = 0;
+
+  do
+    {
+      if (mp->identity_len > 64 || mp->nwi_name_len > 64)
+	{
+	  rv = VNET_API_ERROR_INVALID_VALUE;
+	  break;
```
ivan4th, line 1082: `goto` is often frowned upon, but not in VPP code. So in order to be consistent with the rest of the code, we either need to switch to this `do { ... } while(0)` construct elsewhere, or use `goto` here like in other API handlers (granted, some of them use `out` and some of them use `reply` label, but that IMO is a lesser inconsistency, even though also worth fixing at some point, maybe).
So I'd just use a `reply:` label and a `goto` for this case

ivan4th (changes requested):

upf/upf_api.c
```diff
@@ -1061,6 +1061,78 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+	}
+
+      vec_validate (nwi_name_vec, mp->nwi_name_len);
+      memcpy (nwi_name_vec, nwi_name, mp->nwi_name_len);
+
+      vec_validate (identity_vec, mp->identity_len);
+      memcpy (identity_vec, identity, mp->identity_len);
+
+      rv =
+	vnet_upf_ue_ip_pool_add_del (identity_vec, nwi_name_vec, mp->is_add);
```
ivan4th, line 1092: Note that `vnet_upf_ue_ip_pool_add_del()` doesn't take ownership of vectors you pass to it, instead cloning their contents:
```c
      ueip_pool->identity = vec_dup (identity);
      ueip_pool->nwi_name = vec_dup (nwi_name);
```
which means we've got a leak here and must free `identity_vec` and `nwi_name_vec` afterwards to fix it

demo-exe: see this reply https://github.com/travelping/upg-vpp/pull/340#discussion_r1265969936

upf/upf.c
```diff
@@ -73,6 +73,8 @@ vnet_upf_ue_ip_pool_add_del (u8 * identity, u8 * nwi_name, int is_add)
   upf_ue_ip_pool_info_t *ueip_pool = NULL;
   uword *p;
 
+  identity = vec_dup (identity);
```
ivan4th, line 76: Replace another `vec_dup (identity)` with `identity` in the code below, otherwise we're doing `vec_dup()` twice and leaking memory

ivan4th (changes requested):

upf/upf_cli.c
```diff
@@ -254,7 +254,15 @@ upf_ueip_pool_add_del_command_fn (vlib_main_t * vm,
   nwi_name = upf_name_to_labels (nwi_s);
   vec_free (nwi_s);
 
-  rv = vnet_upf_ue_ip_pool_add_del (name, nwi_name, is_add);
+  if (vec_len (nwi_name) > 64)
+    return clib_error_return (0, "NWI name(encoded) has to fit in 64 bytes");
+  if (vec_len (identity) > 64)
+    return clib_error_return (0, "UE IP pool name has to fit in 64 bytes");
```
ivan4th, line 260: In this case, we need to free these vectors. Also, there's a similar problem with `nwi_s` which existed in the original code. One way to fix it: assign `0` to `name`, `nwi_name`, and `identity` at the beginning of the func, assign `error` val and `goto done` instead of returning immediately upon an error, and at the `done` label, `vec_free()` them. Note that you do not need to check if the vars are `0` or not, as `vec_free()` does the check

upf/upf_cli.c
```diff
@@ -227,18 +227,21 @@ upf_ueip_pool_add_del_command_fn (vlib_main_t * vm,
...
   u8 *nwi_s = 0;
   u8 *nwi_name;
   int rv = 0;
   int is_add = 1;
 
   if (!unformat_user (main_input, unformat_line_input, line_input))
-    return 0;
+    {
+      error = 0;
+      goto done;
```
ivan4th, line 239: This will crash, see comments for #340

ivan4th (approved):
LGTM

PR #340, opened 2023-06-19
[CENNSO-920] feat: binapi for nat pool

demo-exe:


upf/upf_api.c
```diff
@@ -1061,8 +1061,39 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+  u8 *name = mp->names;
+  u8 *nwi_name;
+  ip4_address_t start, end;
+  u16 min_port = UPF_NAT_MIN_PORT;
+  u16 max_port = UPF_NAT_MAX_PORT;
+  u32 port_block_size;
+  u8 is_add = mp->is_add;
 
+  // calculate start of second cstring
+  while (*name++);
```
ivan4th, line 1083: This will fail in an unpredictable way if `name` doesn't happen to contain a zero byte. Please check for this situation

upf/upf_api.c
```diff
@@ -1061,8 +1061,39 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+  u8 *nwi_s_vec = 0;
+  u8 *name = mp->names;
+  u8 *nwi_name;
+  ip4_address_t start, end;
+  u16 min_port = UPF_NAT_MIN_PORT;
+  u16 max_port = UPF_NAT_MAX_PORT;
+  u32 port_block_size;
+  u8 is_add = mp->is_add;
 
+  // calculate start of second cstring
```
ivan4th, line 1082: Please use `/* ... */` style comments for consistency

test/e2e/upg_e2e.go
```diff
@@ -604,6 +604,93 @@ var _ = ginkgo.Describe("CLI debug commands", func() {
 
 // TODO: validate both binapi and CLI against each other
 var _ = ginkgo.Describe("UPG Binary API", func() {
+	ginkgo.Context("for upf nat pool", func() {
+		f := framework.NewDefaultFramework(framework.UPGModeTDF, framework.UPGIPModeV4)
+
+		addPool := func(start string, end string, identity string, name string, isAdd bool) {
+			start_ip, _ := ip_types.ParseIP4Address(start)
+			end_ip, _ := ip_types.ParseIP4Address(end)
```
ivan4th, line 612: Please don't use underscores in Go variable names

test/e2e/upg_e2e.go
```diff
@@ -604,6 +604,93 @@ var _ = ginkgo.Describe("CLI debug commands", func() {
...
+				IsAdd: isAdd,
+
+				MinPort:   2000,
+				MaxPort:   3000,
+				BlockSize: 1,
+				Start:     start_ip,
+				End:       end_ip,
+
+				Nwi:    nwi,
+				NwiLen: uint8(len(nwi)),
```
ivan4th, line 627: NWI length should be calculated automatically by GoVPP, we don't need to specify it here

upf/upf_api.c
```diff
@@ -1061,8 +1061,44 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+
+  mp->min_port = ntohs (mp->min_port);
+  mp->max_port = ntohs (mp->max_port);
+  mp->block_size = ntohl (mp->block_size);
 
+  ASSERT (mp->nwi_len <= 64);
+  vec_validate (nwi_vec, mp->nwi_len);
+  memcpy (nwi_vec, mp->nwi, mp->nwi_len);
+
+  name_vec_len = strlen (mp->name) + 1;
```
ivan4th, line 1085: This may not work correctly if `mp->name` is 64 bytes long, with no space for the zero byte, as `strlen` will get past the end of the field while trying to find the zero byte

upf/upf_api.c
```diff
@@ -1061,8 +1061,44 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+  u8 *name_vec = 0;
+  u8 *nwi_vec = 0;
+  u32 name_vec_len = 0;
+  ip4_address_t start, end;
+
+  mp->min_port = ntohs (mp->min_port);
+  mp->max_port = ntohs (mp->max_port);
+  mp->block_size = ntohl (mp->block_size);
 
+  ASSERT (mp->nwi_len <= 64);
```
ivan4th, line 1081: This may cause an assertion failure due to bad API call parameters, which is not a correct behavior. The proper behavior here would be returning an error from the API call. Verifying such a situation in the E2E test would also be nice.

ivan4th (changes requested):

upf/upf_api.c
```diff
@@ -1061,8 +1061,49 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+
+  mp->min_port = ntohs (mp->min_port);
+  mp->max_port = ntohs (mp->max_port);
+  mp->block_size = ntohl (mp->block_size);
+  do
+    {
+      if (mp->nwi_len > 64 || mp->name_len > 64)
+	{
+	  rv = VNET_API_ERROR_INVALID_VALUE;
+	  break;
```
ivan4th, line 1085: `goto` is often frowned upon, but not in VPP code. So in order to be consistent with the rest of the code, we either need to switch to this `do { ... } while(0)` construct elsewhere, or use `goto` here like in other API handlers (granted, some of them use `out` and some of them use `reply` label, but that IMO is a lesser inconsistency, even though also worth fixing at some point, maybe).
So I'd just use a `reply:` label and a `goto` for this case

upf/upf_api.c
```diff
@@ -1061,8 +1061,49 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+      vec_validate (name_vec, mp->name_len);
+      memcpy (name_vec, mp->name, mp->name_len);
+
+      ip4_address_decode (mp->start, &start);
+      ip4_address_decode (mp->end, &end);
+
+      rv =
+	vnet_upf_nat_pool_add_del (nwi_vec, start, end, name_vec,
+				   mp->block_size, mp->min_port, mp->max_port,
+				   mp->is_add);
```
ivan4th, line 1099: Note that `vnet_upf_nat_pool_add_del()` doesn't take ownership of vectors you pass to it, instead cloning their contents:
```
      nat_pool->name = vec_dup (name);
      nat_pool->network_instance = vec_dup (nwi_name);
```
which means we've got leak here, and must free `nwi_vec` and `name_vec` afterwards to fix it

ivan4th (changes requested):

upf/upf_api.c
```diff
@@ -1061,8 +1061,47 @@ vl_api_upf_tdf_ul_table_add_t_handler (vl_api_upf_tdf_ul_table_add_t * mp)
...
+  memcpy (nwi_vec, mp->nwi, mp->nwi_len);
 
+  name_vec = vec_new (u8, mp->name_len);
+  memcpy (name_vec, mp->name, mp->name_len);
+
+  ip4_address_decode (mp->start, &start);
+  ip4_address_decode (mp->end, &end);
+
+  rv =
+    vnet_upf_nat_pool_add_del (nwi_vec, start, end, name_vec, mp->block_size,
```
demo-exe, line 1097: @ivan4th this function actually takes ownership of those 2 vectors and frees them later on is_add=0 call.
Freeing them causes hashmap to get corrupt

upf/upf.c
```diff
@@ -157,6 +157,8 @@ vnet_upf_nat_pool_add_del (u8 * nwi_name, ip4_address_t start_addr,
   upf_nat_pool_t *nat_pool = NULL;
   uword *p;
 
+  name = vec_dup (name);
```
ivan4th, line 160: There's second `vec_dup (name)` below which needs to be replaced by just `name`, otherwise we've got a leak

upf/upf_api.c
ivan4th: Now it no longer takes ownership but the vectors aren't freed, still

ivan4th (changes requested):

upf/upf_cli.c
```diff
@@ -348,6 +348,9 @@ upf_nat_pool_add_del_command_fn (vlib_main_t * vm,
 				   port_block_size, min_port, max_port,
 				   is_add);
 
+      vec_free (nwi_name);
+      vec_free (name);
```
ivan4th, line 352: We need to free these vectors in **any** case. Also, there's a similar problem with `nwi_s` which existed in the original code. One way to fix it: assign `0` to `name`, `nwi_name`, and `nwi_s` at the beginning of the func, assign `error` val and `goto done` instead of returning immediately upon an error, and at the `done` label, `vec_free()` them. Note that you do not need to check if the vars are `0` or not, as `vec_free()` does the check

ivan4th (changes requested):
Some more leaks...

upf/upf_cli.c
```diff
@@ -306,7 +306,10 @@ upf_nat_pool_add_del_command_fn (vlib_main_t * vm,
   int rv;
 
   if (!unformat_user (main_input, unformat_line_input, line_input))
-    return 0;
+    {
+      error = 0;
+      goto done;
```
ivan4th, line 311: In this particular case, keeping `return 0` is better, as otherwise the code will crash (see below)

upf/upf_cli.c
```diff
@@ -352,9 +355,12 @@ upf_nat_pool_add_del_command_fn (vlib_main_t * vm,
 	error = clib_error_return (0, "Unable to create NAT Pool");
     }
 
+done:
+  vec_free (nwi_name);
+  vec_free (name);
   vec_free (nwi_s);
+  unformat_free (line_input);
```
ivan4th, line 362: This will crash if `unformat_user` returned 0
See
https://github.com/FDio/vpp/blob/2652867480a185ccdeccf9ca9210a1213b82cf26/src/vppinfra/format.h#L155-L161
https://github.com/FDio/vpp/blob/2652867480a185ccdeccf9ca9210a1213b82cf26/src/vppinfra/unformat.c#L398-L407
this happens when an empty input (rest of the command) is encountered
https://github.com/FDio/vpp/blob/2652867480a185ccdeccf9ca9210a1213b82cf26/src/vppinfra/unformat.c#L381-L395

ivan4th (changes requested):

ivan4th (approved):
LGTM

PR #344, opened 2023-06-28
[CENNSO-1030] Revert the new timer wheel implementation for now

ivan4th:
It turned out to have a substantial memory leak, possibly with some timers not being cleaned up during session updates. Reverting for now to have more time for debugging and validation


korroot (approved):

RoadRunnr:
I've state many times that the changed timer wheel had a good chance of having memory leaks. So, this is nothing new or surprising.
Instead of reverting the changes, finding and fixing the missing timer cleanups would be much more helpful.

ivan4th:
@RoadRunnr That's the plan actually, to have the changes merged back as soon as we fix the issues. The revert is needed as a temporary measure due to release timings

PR #351, opened 2023-08-02
[CENNSO-767] feat: added proper errors to packets dropped in upf-ip[46]-forward node

demo-exe:


ivan4th (approved):
LGTM

ivan4th:
I think this is ok, except for one minor detail: I think it would make a little bit more sense to return `true` from `process_urrs()` and `process_qers()` when the packet is NOT to be dropped (b.c. "processed successfully", roughly speaking, it's better when the packets don't have to be dropped ;)

ivan4th (approved):
LGTM

PR #353, opened 2023-08-14
[CENNSO-1149] feat: drop sessions upon receiving "Session context not found" errors

demo-exe:


mgumz:
@ivan4th any objections on this PR?

demo-exe:
@mgumz this is still a draft and not in review, should be ready later today or tommorow, hang tight!

upf/upf_pfcp_server.c
```diff
@@ -369,9 +369,9 @@ upf_pfcp_server_rx_msg (pfcp_msg_t * msg)
 
 	msg->node = req->node;
 
-	pfcp_msg_pool_put (psm, req);
+	upf_pfcp_handle_msg (req, msg);
```
RoadRunnr, line 372: The only real need for passing `req` along is to later extract the session id.

It would be much simpler is you took the `session_index` from req and copied that to `msg`. That would also avoid the need for decoding the request.

The other use case for passing the requests is to ensure that the response message type matches the requests. However, that could be done more generic and for all response types in `upf_pfcp_server_rx_msg` directly.

RoadRunnr (dismissed):

RoadRunnr:
LGTM, I'll leave the final approval to @ivan4th

upf/upf_pfcp_api.c
```diff
@@ -2892,9 +2892,32 @@ handle_session_report_request (pfcp_msg_t * msg, pfcp_decoded_msg_t * dmsg)
...
 handle_session_report_response (pfcp_msg_t * msg, pfcp_decoded_msg_t * dmsg)
 {
+  if (dmsg->session_report_response.response.cause ==
+      PFCP_CAUSE_SESSION_CONTEXT_NOT_FOUND)
+    {
+      // control plane does not know about the session. drop it
+      upf_main_t *gtm = &upf_main;
+      upf_session_t *sess;
+
+      if (!(sess = pool_elt_at_index (gtm->sessions, msg->session_index)))
```
ivan4th, line 2910: `pool_elt_at_index()` contains `ASSERT`, ensuring the item is not free, so in order to avoid a crash you need to check `pool_is_free_index()` beforehand. `pool_elt_at_index` never returns a null pointer.

There's another concern, namely that there may be an edge case with the following sequence of events:
1. UPG sends PFCP Session Report Response to the SMC, but it doesn't arrive immediately
2. The session is deleted by the SMC
3. A new session is created by the SMC and takes up the same index in the pool as a prev session
4. The Session Report Response gets to the SMC and is replied to with Session context not found, as the session has already been deleted
5. A late Session Report response is sent from SMC towards UPG with Session context not found error
6. UPG handling this report will cause the session created in p.3 to be deleted, even though it's fully ok

As a safeguard, we could probably add `u64 seid` field to the `pfcp_msg_t` and compare it to `sess->cp_seid` after we get the session by its index

upf/upf_pfcp_server.c
```diff
@@ -368,6 +368,7 @@ upf_pfcp_server_rx_msg (pfcp_msg_t * msg)
 	upf_pfcp_server_stop_msg_timer (req);
 
 	msg->node = req->node;
+	msg->session_index = req->session_index;
```
ivan4th, line 371: if we add `seid` to `msg`, we also need to copy it here

ivan4th (changes requested):

test/e2e/util/fit.go
```diff
@@ -0,0 +1,49 @@
...
+	if h == nil {
+		return false
+	}
+
+	val, exists := h.faultEnabled[name]
+	if !exists {
+		return false
+	}
+
+	return val
```
ivan4th, line 44: You can just use
```suggestion
	return h.faultEnabled[name]
```
In this case, `false` is returned in case `name` is not present in the `faultEnabled` map.
There's no "uncertainty" associated with such an approach in Go.

test/e2e/util/fit.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func (h *FITHook) EnableFault(name string) {
+	h.faultEnabled[name] = true
```
ivan4th, line 48: You can do this instead:
```suggestion
	if h.faultEnabled == nil {
		h.faultEnabled = make(map[string]bool)
	}
	h.faultEnabled[name] = true
```

This way, you don't need the `NewFITHook()` constructor at all. In Go, it's often preferred not to have a constructor function if you can't do without one.

test/e2e/util/fit.go
```diff
@@ -0,0 +1,49 @@
...
+const (
+	// Faults
+	FaultSessionForgot = "session_forgot"
+)
+
+func NewFITHook() *FITHook {
+	return &FITHook{
+		faultEnabled: make(map[string]bool),
+	}
+}
```
ivan4th, line 32: This constructor function could be eliminated, see the suggestion below

ivan4th (changes requested):
Some mostly cosmetic suggestions related to the `FITHook` (in addition to some previously requested changes in the C part)

ivan4th (approved):
LGTM

PR #355, opened 2023-09-25
[CENNSO-1275] Reduce room for floating point errors in timerwheel

mogaika:
Truncate timerwheel init time and use pow2 clocks per second to basically convert internal time of timerwheel from f64 to i52 by avoiding any floating point rounding.

This reduces timerwheel capacity from ~2985 hours to ~2332 hours

Timerwheel internal time is subject to rounding error and before this patch imprecise calculations resulted in incorrect calculation of clock cycles inside timerwheel implementation.

### DIFF
--- a/upf/upf_pfcp_server.c
+++ b/upf/upf_pfcp_server.c
@@ -36,8 +36,9 @@
 
 #define RESPONSE_TIMEOUT 30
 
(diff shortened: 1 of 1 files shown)


ivan4th (approved):
LGTM

PR #360, opened 2023-11-07
[CENNSO-1370] Remove interface stats increment from upf nodes

mogaika:
They broken (wrong iface_id, missed cache flush), never worked and shouldn't be here in the first place. Since they increase software interface stats and this should be done once in input nodes.


mgumz:
@mogaika which stats/metrics are affected by this change?

mogaika:
> @mogaika which stats/metrics are affected by this change?

None, this is remove of leftover from gtpu plugin which upg was based on

demo-exe (approved):
lgtm but i'd go with title of
'[CENNSO-1370] refactor: cleaup of unused interface stats'

PR #361, opened 2023-11-14
[CENNSO-150] feat: implement MPAS

mogaika:
Implement SMFSet and migration of session to new SMF association from SMFSet when old association is released. And resend in-flight request to new SMF when old SMF goes down to not lose reports.

Session doesn't aware if association setup it's part of is SMFSet or not except for case when association released and session cp_fseid becomes invalid until new SMF updates it, but even this part handled in pfcp messaging. Only major change of existing logic is in association release code, which for SMFSet case doesn't remove sessions, but trying to redistribute sessions across remaining associations in this SMFSet. Other changes mostly bookkeeping.

Directly unrelated change to MPAS:
- Separate up_seid and cp_seid in code and test. Use cp_seid only when we send messages, in other cases use up_seid. Previously UPF used cp_seid as up_seid, but with migration it can be possible that new SMF will set different cp_seid and we wouldn't be able to used it as up_seid since we can't change up_seid
- Add tracking of cp_f_seid to detect cp_f_seid collisions. cp_f_seid ip addresses cached in separated pool to not increase memory usage per session noticeably
- Tracking of sent request per session to resend them when we migrate between SMFs
- Unification of tracking of sessions per association using common linked list implementation
- Do not create spinlock for session if it will not be used (in macro) to decrease memory usage.

### DIFF
--- a/test/e2e/util/fit.go
+++ b/test/e2e/util/fit.go
@@ -23,8 +23,8 @@ type FITHook struct {
 }
 
 const (
-	FaultSessionForgot  FaultType = "session_forgot"
-	FaultReportResponse FaultType = "no_report_response"
+	FaultSessionForgot    FaultType = "session_forgot"
+	FaultNoReportResponse FaultType = "no_report_response"
 
 	// IgnoreHeartbeatRequests makes PFCPConnection ignore incoming
 	// PFCP Heartbeat Requests, thus simulating a faulty CP.
--- a/test/e2e/util/fit.go
+++ b/test/e2e/util/fit.go
@@ -23,7 +23,8 @@ type FITHook struct {
 }
 
 const (
-	FaultSessionForgot FaultType = "session_forgot"
+	FaultSessionForgot  FaultType = "session_forgot"
+	FaultReportResponse FaultType = "no_report_response"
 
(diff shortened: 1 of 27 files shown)


test/e2e/util/fit.go
```diff
@@ -23,7 +23,8 @@ type FITHook struct {
 }
 
 const (
-	FaultSessionForgot FaultType = "session_forgot"
+	FaultSessionForgot  FaultType = "session_forgot"
+	FaultReportResponse FaultType = "no_report_response"
```
demo-exe, line 27:
```suggestion
	FaultNoReportResponse FaultType = "no_report_response"
```

upf/upf_pfcp_server.h
```diff
@@ -73,14 +73,30 @@ typedef struct
   } lcl;
 
   u32 node;
-  u32 session_index;
-  u64 seid;
+
+  struct
+  {
+    /*
+       request can lost session in-flight if session was removed
```
demo-exe, line 80:
```suggestion
       request can loose session in-flight if session was removed
```

demo-exe (approved):
lgtm

PR #371, opened 2023-12-21
style: added .clang-format

demo-exe:
- added `.clang-format`
- `find . -name "*.c" -or -name "*.h" | xargs sed -e 's#/\* \*INDENT-ON\* \*/#/\* clang-format on \*/#g' -e 's#/\* \*INDENT-OFF\* \*/#/\* clang-format off \*/#g' -i`
- `find . -name "*.c" -or -name "*.h" | xargs clang-format-11 -i`
- substituted `hack/checkstyle.sh` with simpler call to clang-format-11 in `Makefile`
- removed `.dir-locals.el` as it is likely to be wrong now
- fixed ci checkstyle to use clang-format


upf/flowtable.c
```diff
@@ -28,19 +28,22 @@
 #if CLIB_DEBUG > 1
 #define upf_debug clib_warning
 #else
-#define upf_debug(...)				\
-  do { } while (0)
+#define upf_debug(...)                                                        \
+  do                                                                          \
+    {                                                                         \
+    }                                                                         \
+  while (0)
```
mogaika, line 35: maybe it makes sense to somehow enable oneliners?

upf/flowtable.c
```diff
@@ -651,17 +649,15 @@ upf_show_flow_timeout_command_fn (vlib_main_t * vm,
 }
 
 /* *INDENT-OFF* */
```
mogaika, line 651: If I not mistaken INDENT-OFF has no use for us then, or no?

upf/flowtable_tcp.h
```diff
@@ -45,12 +45,9 @@ typedef enum tcp_event
 
 /* flow lifetime in seconds */
 static const int tcp_lifetime[TCP_F_STATE_MAX] = {
-  [TCP_F_STATE_SYN] = 15,
-  [TCP_F_STATE_SYNACK] = 60,
-  [TCP_F_STATE_ESTABLISHED] = 299,
-  [TCP_F_STATE_FIN] = 15,
-  [TCP_F_STATE_FINACK] = 3,
-  [TCP_F_STATE_RST] = 6
+  [TCP_F_STATE_SYN] = 15,          [TCP_F_STATE_SYNACK] = 60,
```
mogaika, line 48: If there is possibility to avoid line merge in such case it would be good (maybe with something similar to `// clang-format off`)

upf/pfcp.c
```diff
@@ -1457,16 +1512,12 @@ free_forwarding_policy (void *p)
 }
 
 static char *destination_interface_name[] = {
-  [0] = "Access",
-  [1] = "Core",
-  [2] = "SGi-LAN",
-  [3] = "CP-Function",
-  [4] = "LI Function",
-  [5] = "5G VN internal",
+  [0] = "Access",      [1] = "Core",        [2] = "SGi-LAN",
```
mogaika, line 1515: same

upf/pfcp.c
```diff
@@ -4946,1496 +4949,1384 @@ encode_bbf_up_function_features (void *p, u8 ** vec)
...
+                                               activate_predefined_rules) },
+  [CREATE_PDR_ACTIVATION_TIME] = { .type = PFCP_IE_ACTIVATION_TIME,
+                                   .offset = offsetof (pfcp_create_pdr_t,
+                                                       activation_time) },
+  [CREATE_PDR_DEACTIVATION_TIME] = { .type = PFCP_IE_DEACTIVATION_TIME,
+                                     .offset = offsetof (pfcp_create_pdr_t,
+                                                         deactivation_time) },
+  [CREATE_PDR_MAR_ID] = { .type = PFCP_IE_MAR_ID,
+                          .offset = offsetof (pfcp_create_pdr_t, mar_id) },
+};
```
mogaika, line 4995: sad, became unreadable. Maybe we should use `clang format off` in this file

upf/pfcp.c
```diff
@@ -4946,1496 +4949,1384 @@ encode_bbf_up_function_features (void *p, u8 ** vec)
...
+  [TP_ERROR_REPORT_TP_FILE_NAME] = { .type = PFCP_IE_TP_FILE_NAME,
+                                     .vendor = VENDOR_TRAVELPING,
+                                     .offset = offsetof (
+                                       pfcp_tp_error_report_t, file_name) },
+  [TP_ERROR_REPORT_TP_LINE_NUMBER] = { .type = PFCP_IE_TP_LINE_NUMBER,
+                                       .vendor = VENDOR_TRAVELPING,
+                                       .offset =
+                                         offsetof (pfcp_tp_error_report_t,
+                                                   line_number) },
+};
```
mogaika, line 6288: same

upf/unittest.c
```diff
@@ -200,96 +209,107 @@ ip_app_test_v6 (void)
...
+                                           0x20010db800120000),
+                                         clib_host_to_net_u64 (
+                                           0x0000000000000002),
+                                       } };
+  ip46_address_t ip_2001_db8_12__3 = { .ip6.as_u64 = {
+                                         clib_host_to_net_u64 (
+                                           0x20010db800120000),
+                                         clib_host_to_net_u64 (
+                                           0x0000000000000003),
+                                       } };
```
mogaika, line 235: same

upf/upf_flow_node.c
```diff
@@ -41,36 +44,32 @@ typedef struct
...
-	      format_white_space, indent,
-	      format_ip4_header, t->packet_data, sizeof (t->packet_data));
+              "upf_session%d up-seid 0x%016llx\n"
+              "%UFlowInfo - sw_if_index %d, next_index = %d\n%U%U\n%U%U\n%U%U",
+              t->session_index, t->up_seid, format_white_space, indent,
+              t->sw_if_index, t->next_index, format_white_space, indent,
+              format_flow_key, &t->key, format_white_space, indent,
+              format_hex_bytes, &t->key, sizeof (t->key), format_white_space,
+              indent, format_ip4_header, t->packet_data,
+              sizeof (t->packet_data));
```
mogaika, line 65: original intent was to separate format function on each line, now it is lost

upf/upf_ipfix_templates.h
```diff
@@ -32,250 +32,220 @@
...
+#define IPFIX_FIELD_INTERFACE_NAME(F)                                         \
+  F (interfaceName, 65535, IPFIX_VALUE_STRING, info->interface_name,          \
+     vec_len (info->interface_name), 1)
+#define IPFIX_FIELD_OBSERVATION_DOMAIN_NAME(F)                                \
+  F (observationDomainName, 65535, IPFIX_VALUE_STRING,                        \
+     info->observation_domain_name, vec_len (info->observation_domain_name),  \
+     1)
+#define IPFIX_FIELD_OBSERVATION_POINT_ID(F)                                   \
+  F (observationPointId, 8, IPFIX_VALUE_U64, info->observation_point_id,      \
+     sizeof (u64), 1)
```
mogaika, line 217: original intent is lost

upf/upf_pfcp.c
```diff
@@ -2815,140 +2801,102 @@ process_qers (vlib_main_t * vm, upf_session_t * sess,
...
-
-static const char *urr_status_flags[] = {
-  "OVER QUOTA",
-  "AFTER MONITORING TIME",
-  "REPORTED",
-  NULL
-};
+static const char *urr_status_flags[] = { "OVER QUOTA",
+                                          "AFTER MONITORING TIME", "REPORTED",
+                                          NULL };
```
mogaika, line 2850: formatting is lost

upf/upf_pfcp_session_server.c
```diff
@@ -423,50 +417,49 @@ pfcp_session_server_main_init (vlib_main_t * vm)
...
-* Local Variables:
-* eval: (c-set-style "gnu")
-* End:
-*/
+ * fd.io coding-style-patch-verification: ON
+ *
+ * Local Variables:
+ * eval: (c-set-style "gnu")
+ * End:
+ */
```
mogaika, line 465: do we need this footers with clang-format?

mogaika (changes requested):
Overall good, but I see problems with static arrays. It would be great to disable folding/compression of arrays/structures, if possible.

And same thing for `format` arguments, we want to keep them on separated lines, so maybe we can disable compression/folding of lines on function/macro calls. Usually it's better to manage lines ourselves in this cases.

If it's not possible, then I think `// clang-format off` has to be used in some places

upf/flowtable_tcp.h
demo-exe: So there is no automatic way at the moment https://github.com/llvm/llvm-project/issues/61560
we can go with clang-format off but personally this does not bother me

mogaika (approved):

PR #379, opened 2024-01-14
[CENNSO-1566] test: fix ipfix test reporting interval timing

mogaika:
This test requires at least 4 reports and UDP traffic generation wasn't running long enough to supply them


demo-exe (approved):
so to uphold conventional commits title should be something like
`[CENNSO-1566] fix: ipfix test reporting interval timing`

PR #381, opened 2024-01-29
feat: generalized buildenv EXTRA_DIR -> EXTRA_ARGS

demo-exe:
this allows mounting x socket inside buildenv for shared clipboard and other things


mgumz:
@demo-exe x11? why?

demo-exe:
shared clipboard support :)

mgumz:
@demo-exe but … what … for?

mogaika (approved):

PR #382, opened 2024-02-01
refactor: flowtable direction logic and other fixes

mogaika:
Functional change:
- Instead of enabling flowtable timers expiration process only in tests, create flowtable timer node which enabled always. Implemented using process node which triggers interrupts in input node which does timer wheel rotation. For now this is only done as backup mechanism for low-traffic situation and flow timers expected to expire mostly in flow node, so wheel rotation frequency is low. To switch entirely to timer expiration node granularity of time has to be increased to spread timers over time and to not burst them every second.

Changes related to flowtable, but not affecting behavior:
- In flow structure do not use reversed order from flow hashkey. This eliminates tons of xoring of direction with flow reverse mark. This makes code much clearer. Matching in hashtable still done with "ordered" key which order depends on ip addresses. Flow internal ip and port fields swapped if needed on flow creation and flow removal. Access to flow fields can done directly with direction type without additional xoring.
- Separate per direction fields of flow to separated struct which to have better logical separation. There is no advantage from performance side to keep it as SoA instead of AoS, since fields used in random order in hot-path and even if performance difference exists, it's in margin of error.
- In some places replace FT_ORIGIN/FT_REVERS with more meaningful enums for different operations.
- Remove flow event system and replace it with explicit calls to event handlers.
- Separate `upf_buffer_opaque_t` in own header so we could solve dependency and use it in flowtable code.

Unrelated changes:
- Remove `"upf-ipX-proxy-server-no-conn-output"` nodes since they make no sense in newer VPP versions (as was mentioned in comments), and also there was a bug where `direction` variable have been incorrectly reused between packets in one vector.
- Disable and deprecate "pooling" mode for pfcp. UPG supports interrupts without additional configuration if all VPP nodes are in interrupt mode. Except in vpp v22.10 there is a bug which prevents enabling interrupt mode, and it's fixed in vpp 5f30518.
- Fix vpp hang when looping of proxy traffic happens if PDI/FAR configured wrongly by dealing with TTL.
- Simplify optimization possibility of acl matching and do not match spare fields in case of ipv4.


upf/flowtable.c
```diff
@@ -322,17 +310,16 @@ format_flow_key (u8 *s, va_list *args)
   flow_key_t *key = va_arg (*args, flow_key_t *);
 
   return format (s, "proto 0x%x, %U:%u <-> %U:%u, seid 0x%016llx", key->proto,
-                 format_ip46_address, &key->ip[FT_ORIGIN], IP46_TYPE_ANY,
-                 clib_net_to_host_u16 (key->port[FT_ORIGIN]),
-                 format_ip46_address, &key->ip[FT_REVERSE], IP46_TYPE_ANY,
-                 clib_net_to_host_u16 (key->port[FT_REVERSE]), key->up_seid);
+                 format_ip46_address, &key->ip[FTK_EL_SRC], IP46_TYPE_ANY,
```
mogaika, line 313: add direction

upf/upf_adf.c
```diff
@@ -296,9 +296,10 @@ app_scan_for_uri (u8 *uri, flow_entry_t *flow, struct rules *active,
         {
           const ip46_address_t *addr;
 
-          addr =
-            &flow->key.ip[direction ^ flow->is_reverse ^
-                          !!(pdr->pdi.ue_addr.flags & PFCP_UE_IP_ADDRESS_SD)];
+          flow_direction_op_t pdi_dir_op =
+            (pdr->pdi.ue_addr.flags & PFCP_UE_IP_ADDRESS_SD) ? FTD_OP_FLIP :
+                                                               FTD_OP_SAME;
+          addr = &flow->key.ip[direction ^ pdi_dir_op];
```
mogaika, line 302: add _SRC_

demo-exe (approved):

demo-exe:
maybe name to 'refactor: flowtable direction logic and other fixes'

PR #388, opened 2024-02-16
[CENNSO-1762] fix: reliable out of volume quota handling

mogaika:
Instead of trying to catch moment when traffic counter is stepped over quota/threshold, track quota/threshold statuses to more reliably catch case when we out of quota/threshold after URR update.

### DIFF
--- a/upf/upf.h
+++ b/upf/upf.h
@@ -596,6 +596,7 @@ typedef struct
 #define URR_OVER_QUOTA            BIT (0)
 #define URR_AFTER_MONITORING_TIME BIT (1)
 #define URR_REPORTED              BIT (2)
+#define URR_OVER_VOLUME_THRESHOLD BIT (3)
 
   u8 update_flags;
 #define PFCP_URR_UPDATE_VOLUME_QUOTA        BIT (0)
--- a/upf/upf_pfcp.c
+++ b/upf/upf_pfcp.c
@@ -2485,13 +2485,11 @@ urr_increment_and_check_counter (u64 *packets, u64 *bytes, u64 *consumed,
 {
   int r = URR_OK;
 
-  if (quota != 0 &&
-      PREDICT_FALSE (*consumed < quota && *consumed + n_bytes >= quota))
+  if (quota != 0 && PREDICT_FALSE (*consumed + n_bytes >= quota))
     r |= URR_QUOTA_EXHAUSTED;
   *consumed += n_bytes;
 
-  if (threshold != 0 &&
-      PREDICT_FALSE (*bytes < threshold && *bytes + n_bytes >= threshold))
+  if (threshold != 0 && PREDICT_FALSE (*bytes + n_bytes >= threshold))
     r |= URR_THRESHOLD_REACHED;
   *bytes += n_bytes;
 
@@ -2623,7 +2621,7 @@ display_packet_for_urr (vlib_main_t *vm, vlib_buffer_t *b,
 /**
  * @brief Function to process URRs.
  *
- * @return true if the packet should be dropped, false otherwise
+ * @return true if the packet consumed quota and is allowed to be forwarded
  */
 bool
 process_urrs (vlib_main_t *vm, upf_session_t *sess, const char *node_name,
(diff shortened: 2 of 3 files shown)


upf/upf_pfcp.c
```diff
@@ -2417,22 +2417,27 @@ pfcp_lookup_cp_f_seid (pfcp_ie_f_seid_t *f_seid)
 }
 
 static int
-urr_increment_and_check_counter (u64 *packets, u64 *bytes, u64 *consumed,
-                                 u64 threshold, u64 quota, u64 n_bytes)
+urr_increment_and_check_counter (urr_volume_t *volume, u8 urr_status,
```
demo-exe, line 2420: typedef enum this u8 urr_status?

upf/upf_pfcp.c
```diff
@@ -2560,7 +2565,7 @@ display_packet_for_urr (vlib_main_t *vm, vlib_buffer_t *b,
 /**
  * @brief Function to process URRs.
  *
- * @return true if the packet should be dropped, false otherwise
+ * @return true if the packet consumed quota and allowed to be forwarded
```
demo-exe, line 2568:
```suggestion
 * @return true if the packet consumed quota and is allowed to be forwarded
```

demo-exe (changes requested):

RoadRunnr:
The meat of this PR seems to be the recording of OVER_QUOTA and THRESHOLD_REACHED state in the URR. Tat seems resonable. But almost everything else looks to be to be just noise that is not doing anything useful.

> Use union of array to access fields of urr_volume_counter_t in more straightforward way

I'm not convicend that this actually an improvement. The static function accessing the URRs will most likes be inlined (or can be force to be inline) in which case the compiler is high likely to generate almost the same or at least very similar code for the old and the new version.
Changing the datastructur from an plain struct to an union where an array aliases the structure member makes this more error prone without adding any benefit.

> Rename urr_measure_t to urr_volume_measure_t

Seems reasonable, but for the sake of reviewing this, it would be much easier to have a separate commit within the PR for the rename only.

upf/upf_pfcp.c
mogaika: agree, but let's postpone it to keep scope for bugfix smaller

upf/upf.h
```diff
@@ -534,27 +534,39 @@ typedef enum
...
-  u64 total;
-} urr_counter_t;
+  struct
+  {
+    u64 ul;
+    u64 dl;
+    u64 total;
+  };
+  u64 as_array[URR_VOLUME_N_COUNTERS];
+} urr_volume_counter_t;
```
RoadRunnr, line 556: Not sure if you haev seen my comment, so let me repeat it as inline comment on the relevant lines.

I'm not convicend that this actually an improvement. The static function accessing the URRs will most likes be inlined (or can be force to be inline) in which case the compiler is highly likely to generate almost the same or at least very similar code for the old and the new version.
Changing the datastructur from an plain struct to an union where an array aliases the structure member makes this more error prone without adding any benefit.

mogaika: It's not improvement for compiler, but for reader. Mostly to get rid of `urr_incr_and_check` macro which does strange access to fields and make it clear which fields are accessed in urr_increment_and_check_counter. Also to get rid of offsetof in format.

RoadRunnr: I see it exactly the other way arround, the change makes it IMHO harder for the reader to understand the code. 

Besides the style issue which could be down personal prefences, I still maintain that aliasing the field members through the union with the array is a bad idea. Constructs like that should only be used if there is absolutely not other god way to achive the same result.
If you insist on using an array, then use only the array and get rid of the struct.

The use of offsetof is valid and well known C syntax. I have no idea what would be a problem with that.

mogaika: Oh, okay. @demo-exe has similar position. So I will cut this commit to minimal change and prepare separate one with more types and use of arrays everywhere.

demo-exe (approved):

PR #389, opened 2024-02-18
[CENNSO-1767] feat: pfcp requests metrics 

demo-exe:


upf/upf.h
```diff
@@ -519,21 +519,118 @@ typedef enum
   UPF_TIMERS_MISSED = 8,
   UPF_IPFIX_RECORDS_SENT = 9,
   UPF_IPFIX_MESSAGES_SENT = 10,
-  UPF_N_COUNTERS = 11,
+  // total of PFCP messages received - corrupted
+  UPF_PFCP_RECEIVED_CORRUPTED = 11,
+  // PFCP request received
+  // parts of program depend on order - OK then ERROR for the same type
```
mogaika, line 525: maybe
```suggestion
  // message _ERROR counter always should be after _OK, since arithmetic used to access it
```

upf/upf.h
```diff
@@ -519,21 +519,118 @@ typedef enum
   UPF_TIMERS_MISSED = 8,
   UPF_IPFIX_RECORDS_SENT = 9,
   UPF_IPFIX_MESSAGES_SENT = 10,
-  UPF_N_COUNTERS = 11,
+  // total of PFCP messages received - corrupted
+  UPF_PFCP_RECEIVED_CORRUPTED = 11,
+  // PFCP request received
```
mogaika, line 524: not only requests, maybe messages?

upf/upf_pfcp_api.c
```diff
@@ -3101,12 +3102,14 @@ upf_pfcp_handle_msg (pfcp_msg_t *msg)
...
       /* not enough info in the message to produce any meaningful reply */
       upf_debug ("PFCP: broken message");
+      upf_increment_counter (UPF_PFCP_RECEIVED_CORRUPTED, 0, 1);
       return -1;
     }
 
   if (r != 0) /* if cause != 0 */
     {
       upf_debug ("PFCP: error response %d", r);
+      upf_increment_counter (UPF_PFCP_RECEIVED_CORRUPTED, 0, 1);
```
mogaika, line 3112: feels like we could use `UPF_PFCP_***_ERROR` here as well, since type is known and error response is sent

mogaika (changes requested):

upf/upf.h
```diff
@@ -519,21 +519,118 @@ typedef enum
...
+  UPF_PFCP_ASSOCIATION_UPDATE_RESPONSE_ERROR = 27,
+  UPF_PFCP_ASSOCIATION_RELEASE_REQUEST_OK = 28,
+  UPF_PFCP_ASSOCIATION_RELEASE_REQUEST_ERROR = 29,
+  UPF_PFCP_ASSOCIATION_RELEASE_RESPONSE_OK = 30,
+  UPF_PFCP_ASSOCIATION_RELEASE_RESPONSE_ERROR = 31,
+  // PFCP but sent the other way
+  UPF_PFCP_NODE_REPORT_REQUEST_OK = 32,
+  UPF_PFCP_NODE_REPORT_REQUEST_ERROR = 33,
+  UPF_PFCP_NODE_REPORT_RESPONSE_OK = 34,
+  UPF_PFCP_NODE_REPORT_RESPONSE_ERROR = 35,
```
mogaika, line 550: btw, do we want to create counters for all messages, unsupported as well?

upf/upf.h
```diff
@@ -519,21 +519,117 @@ typedef enum
...
   UPF_IPFIX_MESSAGES_SENT = 10,
-  UPF_N_COUNTERS = 11,
+  // total of PFCP messages received - corrupted
+  UPF_PFCP_RECEIVED_CORRUPTED = 11,
+  // PFCP messages received
+  // _ERROR counter should always follow _OK counter
+  UPF_PFCP_RX_HB_REQUEST_OK = 12,
+  UPF_PFCP_RX_HB_REQUEST_ERROR = 13,
+  UPF_PFCP_RX_HB_RESPONSE_OK = 14,
+  UPF_PFCP_RX_HB_RESPONSE_ERROR = 15,
```
mogaika, line 529: what I meant with RX is that each counter may have this prefix to clarify that counter is for received message. So either all, either none

mogaika (changes requested):

upf/upf.h
demo-exe: huh thought that HB could be sent 2 ways, my bad
Id go without those prefixes

